PR #3, opened 2018-07-26
Add an option to send a http GET or POST after config map change.

b13n1u:
I have added possibility to send a simple HTTP request to an URL specified via env vars. 
It is very useful when a service provides an reload endpoint on the API (For example prometheus).
The code is not very sophisticated.  If you don't provide a request method it uses GET as default. 
If you don't provide an URL it does nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # What?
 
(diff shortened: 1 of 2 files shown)


Tim-Smyth:
Hello @b13n1u,
Thank you for your code contribution and sorry for the delay.
Tim

PR #4, opened 2018-07-30
Use NAMESPACE env variable to search on a specific namespace

### LINKED ISSUE #2: Search only in current or specific namespace
Hi,
would it maybe be possible to make the sidecar configurable to search only in a specific or the actual namespace?
In our setup we use namespaces to separate teams and so we don't have access to search in all namespaces.. which is btw not neccessary in our usecase.. so simply set the namespace where the sidecar is looking for config maps would already fit our needs.

Whats you opinion on such a change?
Thanks

eloo:
This will fix #2 
@Tim-Smyth what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,8 @@ Example for a simple deployment can be found in `example.yaml`. Depending on the
   - description: Folder where the files should be placed
   - required: true
   - type: string
+
+- `NAMESPACE`
+  - description: If specified, the sidecar will search for config-maps inside this namespace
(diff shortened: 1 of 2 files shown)


Tim-Smyth:
Looks good @eloo, could you please extend the Readme too?

eloo:
of course ;) 
will do it soon

eloo:
done

Tim-Smyth:
Thank you very much @eloo

Issue #5, opened 2018-08-08
Sidecar crashes with IncompleteRead


clalbus:
**Sidecar version:** Issue has been observed on 0.0.3 and 0.0.5
**Kubernetes version:** 1.10.6 (AKS)

I am running Grafana (using helm stable release) with sidecar for datasources/dashboards ([relevant docs](https://github.com/helm/charts/tree/master/stable/grafana#sidecar-for-dashboards))

After some time sidecar starts to crash (in a loop) with the following errors from the logs:

```
Starting config map collector
Config for cluster api loaded...
Working on configmap kube-system/xxx-grafana.v2
Working on configmap kube-system/xxx-grafana.v4
Working on configmap kube-system/xxx-grafana.v5
Working on configmap kube-system/xxx-grafana.v6
Working on configmap kube-system/addon-http-application-routing-tcp-services
Working on configmap kube-system/addon-http-application-routing-udp-services
Working on configmap kube-system/dashboards.v2
Working on configmap kube-system/xxx-grafana.v1
Working on configmap default/xxx-grafana-config-dashboards
Working on configmap kube-system/addon-http-application-routing-nginx-configuration
Working on configmap kube-system/omsagent-rs-config
Working on configmap kube-system/azureproxy-nginx
Working on configmap default/xxx-prometheus-server
Working on configmap kube-system/api.v1
Working on configmap kube-system/dashboards.v7
Working on configmap kube-system/dashboards.v1
Working on configmap kube-system/dashboards.v5
Working on configmap kube-system/image-storage.v1
Working on configmap kube-system/kubedns-kubecfg
Working on configmap kube-system/xxx-grafana.v3
Working on configmap default/api-grafana-dashboard
Configmap with label found
File in configmap api-dashboard.json ADDED
Working on configmap default/xxx-grafana
Working on configmap kube-system/azureproxy-config
Working on configmap kube-system/dashboards.v6
Working on configmap kube-system/xxx-prometheus.v1
Working on configmap kube-system/render.v1
Working on configmap default/prometheus-grafana-datasource
Working on configmap default/render-grafana-dashboard
Configmap with label found
File in configmap render-dashboard.json ADDED
Working on configmap default/xxx-prometheus-alertmanager
Working on configmap kube-system/dashboards.v3
Working on configmap kube-system/dashboards.v4
Working on configmap kube-system/heapster-config
Working on configmap default/image-storage-grafana-dashboard
Configmap with label found
[...]

PR #6, opened 2018-08-20
Don't crash on exceptions

eloo:
Improve error handling related to #5

### DIFF
--- a/sidecar/sidecar.py
+++ b/sidecar/sidecar.py
@@ -89,8 +89,11 @@ def main():
 
     config.load_incluster_config()
     print("Config for cluster api loaded...")
-    watchForChanges(label, targetFolder, url, method, payload)
-
+    while True:
+        try:
+            watchForChanges(label, targetFolder, url, method, payload)
+        except:
+            print("Unexpected error:", sys.exc_info()[0])
 
 if __name__ == '__main__':
     main()
--- a/sidecar/sidecar.py
+++ b/sidecar/sidecar.py
@@ -1,4 +1,6 @@
 from kubernetes import client, config, watch
+from kubernetes.client.rest import ApiException
+
 import os
 import sys
 import requests
@@ -92,8 +94,11 @@ def main():
     while True:
         try:
             watchForChanges(label, targetFolder, url, method, payload)
+        except ApiException as e:
+            print("Exception when calling kubernetes: %s\n" % e)
         except:
             print("Unexpected error:", sys.exc_info()[0])
+            break
 
 if __name__ == '__main__':
     main()


Tim-Smyth:
With this PR we would take to error handling away from k8s. For example while testing I worked with a dummy notification address and the container didn't fail as expected but keep retrying.

eloo:
@Tim-Smyth ah yes thats right.
but at the moment the container is also not crashing everytime but stops doing anything.. 
so i'm facing not updated files but get restarts for the container.. 
and with this PR i simple want to keep the functionality alive

Tim-Smyth:
If your PR should fix that issue, there should be an exception thrown if he hangs right? And if it does the container should restart anyway? If he hangs without throwing an exception we have to find an different approach to solve it.
About the configmap being invalid: we just use fields that are mandatory/controlled by k8s anyway, so we should not run into problems there.

eloo:
@Tim-Smyth but why should we keep the container restarting for a simple exception? this consume way more time than simply "ok there was an error i will log it and keep going on"

i really didn't see any advantage of restarting the whole container

Tim-Smyth:
The advantage I see is that k8s will make sure that the container doesn't misbehave: If the container fails a few times the backoff strategy will kick in and reduce the "load" on the cluster.
While playing around with the sidecar I configured a dummy url for the notifications, which resulted in repeated restarts of the watchForChanges function, resulting in unnecessary load.

Tim-Smyth:
@eloo You are right about the configuration, but its just an example. With your PR we just broadly cover all exceptions. What if a the permissions of the sidecar's user are modified and now we always get an exception that we are not allowed to access the k8s api? The pod would just try accessing the api as fast as he can without the operator noticing it (or its malfunction).

eloo:
yes that true.. okay, i will have a look if i can be more specific about the exception i'm catching.

ahma:
Hi 
We had a similar problem with the AKS (Azure), and OCI (Oracle). 
I understand your opinion @Tim-Smyth and why you don't want to merge the fix like this, but in this case, I don't have a better idea. 
I made a working solution based on @eloo 's  idea. 
code:  https://github.com/banzaicloud/k8s-sidecar/pull/1/files
Image: banzaicloud/k8s-sidecar:0.0.7
I hope it will be helpful for someone.

MihaiAnei:
Coming from the above: https://github.com/helm/charts/issues/9136

Did you guys know about this: https://github.com/kubernetes-client/python-base/issues/57 ?

Tim-Smyth:
This issue is now covered with PR https://github.com/kiwigrid/k8s-sidecar/pull/15
Hope the new version 0.0.12 solves your issues. Thank you for your valuable input :+1: 
If there are still problems please let me know

PR #7, opened 2018-08-24
feat: limit configmap search to current namespace

giacomoguiulfo:
This will limit the `ConfigMap` watching to the current namespace in which the sidecar is running by `default`. It still allows for a namespace to be specified or `ALL` to watch `ConfigMaps` in all namespaces.


giacomoguiulfo:
@Tim-Smyth

PR #12, opened 2018-11-30
Added list feature for sidecar

xuxinkun:
Sometimes, We will just expect that this container used as init container to collect the configmaps. So, the container should exit after list the configmaps needed.


Tim-Smyth:
HI @xuxinkun, thank you for your PR i really like this feature :+1:

PR #17, opened 2019-02-08
Add requirements.txt file and upgrade deps

aslafy-z:
- Lint Dockerfile
- Update python version
- Add requirements.txt
- Pin all deps and update to latest


Tim-Smyth:
I would be still interested in this PR, if you could be so kind and adjust to the current master

aslafy-z:
I just rebased the PR on the current master.

PR #22, opened 2019-02-21
Base64 Data Decoding Support

daviddyball:
Added support for `.base64` filename suffix to allow decoding base64 data.

For us this gets around some terrible escaping issues we're seeing with our CICD platform (Spinnaker). Instead of escaping all of our `$variable`s in a dashboard to be `${"$variable"}` (which makes it unreadable and impossible to lint), so now we just base64 our JSON and save that in the `ConfigMap`

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ Example for a simple deployment can be found in `example.yaml`. Depending on the
 
 If the filename ends with `.url` suffix, the content will be processed as an URL the target file will be downloaded and used as the content file.
 
+If the filename ends with `.base64` suffix, the content will be base64 decoded and persisted to disk.
+
 ## Configuration Environment Variables
 
 - `LABEL` 
@@ -69,9 +71,3 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - description: Set to true to skip tls verification for kube api calls
   - required: false
   - type: boolean
-
-## Annotations
-
-| Annotation | Description |
-|------------|-------------|
-| `k8s-sidecar.kiwigrid/base64-data` | Tells the sidecar to treat the values in your `ConfigMap` as base64 encoded and saves the decoded output into the destination file |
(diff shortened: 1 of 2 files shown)


Tim-Smyth:
Hi @daviddyball ,
Thank you for your PR. Can you please adjust to the current master and update your PR. Then it should be merged fast. Thanks

Tim-Smyth:
Please update if you are still interested in the PR

migueleliasweb:
Hi, is there any workarounds for this nowadays? Ping @pulledtim 

I would like to mount binary-like files to a container using the `k8s-sidecar`. The problem is that I must b64 encode the files before creating the `ConfigMaps`.

Otherwise, this PR would still be awesome to merge as it allows a whole new suite of possible integrations that can't work with just plain strings in `ConfigMaps`.

jekkel:
Hi @migueleliasweb 

if you're still interested, then please re-open and update the PR.

migueleliasweb:
Yes, sir!

migueleliasweb:
Digging a bit deeper.... 

I've found that the support for base64'd files is actually present but the issue seems to be that it only happens when `binaryData` is used instead of `data`.

Am I seeing this right? https://github.com/kiwigrid/k8s-sidecar/blob/master/src/resources.py#L221

migueleliasweb:
Maybe the issue here is just documenting this feature and making it clear in the `README.md`.

Let me know your thoughts and if I can give it a go on documenting this.

migueleliasweb:
I just tested the functionality and it seems to be working just fine.

Basically I've used a `ConfigMap` with `binaryData` defined to a base64 value and it was decoded before placing the file in the mount.

I guess we're just missing the docs ;).

jekkel:
Indeed... Config maps and secrets can hold binary data by the k8s API and hence we support that.

migueleliasweb:
Sure thing. I should raise a PR soon.

migueleliasweb:
Please have a look: https://github.com/kiwigrid/k8s-sidecar/pull/249

PR #15, opened 2019-02-08
Allow overriding target directory with an annotation

mmiller1:
This PR adds functionality to allow a user to override the default target directory per configmap based on an annotation. I also ported the exception handling implemented in https://github.com/banzaicloud/k8s-sidecar/pull/1/files to prevent excessive container restarts

PR #25, opened 2019-03-08
bugfix: fixed destFolder

jwenz723:
#15 introduced a bug that caused the directory set in the `k8s-sidecar-target-directory` annotation to be used on ConfigMaps that follow.

Anyone interested in testing the fix can use the docker image `jwenz723/k8s-sidecar:0.0.13`

See my comment here: https://github.com/kiwigrid/k8s-sidecar/pull/15#issuecomment-471033354


Tim-Smyth:
Thanks for your fix :+1:

Issue #31, opened 2019-05-11
Offer for help / Feature request: Read/Watch secrets


fr-ser:
Hello,

I wanted to implemented some functionality in an upstream project (grafana helm chart) and was looking for the ability to watch secrets as well as configmaps.

Currently this project does not seem to support that, but I would appreciate this functionality very much. I can also offer to implement it, in case this project is still active and this functionality fits into the scheme of things.

So... Want me to take a crack at enabling secret watching? I think this should not be a breaking change.

Tim-Smyth:
Hi Sergej,
Thank you for your offer. Our team doesn't need that functionality at the moment but that should not stop you. If you are willing to implement it, we would be happy to review and merge it.

fr-ser:
Would it be alright if I add some unit tests with pytest to this project?

Tim-Smyth:
Unit tests would be awesome. Once I am back from vacation I want to set up CI pipeline too.

fr-ser:
Alright. Once the initial branch is merged I would take a stab at adding those then :+1:

Tim-Smyth:
Sorry that PR took so long, don't really have an excuse. I will do my best to get the next one fast done. Thank you very much for your contribution :+1:

PR #32, opened 2019-05-16
Allow secrets

fr-ser:
PR relating to [issue 31](https://github.com/kiwigrid/k8s-sidecar/issues/31).

The first commit was simple refactoring according to flake8.

The second commit touched a lot of files as the single file approach would have been to messy in my opinion. I separated the logic into three files.

There was also a little problem with running two endless iterators in python, which I solved using Threads (only when we really want to watch secrets and configmaps both)

### DIFF
--- a/sidecar/resources.py
+++ b/sidecar/resources.py
@@ -1,6 +1,7 @@
 import base64
 import os
-from threading import Thread
+from multiprocessing import Process
+from time import sleep
 
 from kubernetes import client, watch
 from kubernetes.client.rest import ApiException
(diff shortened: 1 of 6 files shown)


sidecar/resources.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def watchForChanges(label, targetFolder, url, method, payload,
+                    current, folderAnnotation, resources):
+
+    if len(resources) == 2:
+        Thread(target=_watch_resource_loop,
+               args=(label, targetFolder, url, method, payload,
+                     current, folderAnnotation, resources[1], True)
+               ).start()
```
Tim-Smyth: Will an exception in the thread be raised to the main thread too? If not we would not have a change of restarting if that thread fails or?

fr-ser: Good point. That is why I put the error check inside the loop.

But if we get a 500 error it is raised and dies... I will think about it this evening a bit and put in a change (I am thinking about checking in the main thread if the loop thread is still running and raising there otherwise...)

fr-ser:
Running one loop in the main thread and another in a separate one complicated things...

Now I handle it consistently and run each loop in its own thread (easier to terminate).
The main process checks each 5 seconds if the loops are still running

fr-ser:
Anything outstanding I can do to help merge this PR?

PR #47, opened 2019-09-18
adding non-root user for sidecar container

vsamidurai:
- Adding non-privileged user for container.


vsamidurai:
Can this be merged with master ? Please let me know, if there is any issue.

PR #49, opened 2019-10-07
Only remove the local file if it's no longer in use by any namespace

### LINKED ISSUE #44: Configmaps containing files deployed to many namespaces causes premature delete
We are installing a configmap (containing a label that the sidecar picks up) to 2 different namespaces (namespaceA and namespaceB). WHEN the configmap is installed to namespaceA, the file in the configmap is added correctly to the specified directory. When the configmap is installed to namespaceB, the sidecar detects that the dashboard contained in the configmap is already discovered and is not added a second time... ALL is good until this point. When namespaceB is wiped of all of its resources the sidecar detects that the configmap is deleted, the file is deleted and is no longer available to namespaceA, despite the namespace persisting and containing the configmap.

The sidecar doesn't keep track that multiple deployments contain the configmap. Hence application that use the file contained in the configmap installed in namespace A are failing.

This issue has been discovered since we are installing a helm chart containing the configmap to a namespace that persists and to another namespace that is just present while tests are being run against it.

**What you expected to happen:**

The sidecar would detect and keep track of which releases contain configmaps which use the file, and only remove the file when all releases using that file are deleted

**How to reproduce it (as minimally and precisely as possible):**

1. Install configmap describing file, with label detectable by the sidecar twice (two seperate releases)
2. Delete one of the releases
3. Check specified directory where files in configmap resources are to be written to.

**Potential Fix**

Store a map of {metadata.namespace}/{metadata.name} to file for each resource with the specified label, and only run delete if ALL entries mapped to that file are deleted

https://github.com/kiwigrid/k8s-sidecar/blob/dd39aec87748c5e62a20d51f831658af96e94f37/sidecar/resources.py#L110-L114

pnordh:
Keep track of where a file has been deployed (namespace) and only delete it if all configmaps/secret resources that was using it has also been removed.

fixes #44

### DIFF
--- a/sidecar/resources.py
+++ b/sidecar/resources.py
@@ -5,6 +5,7 @@ import os
 import sys
 import signal
 
+from collections import defaultdict
 from multiprocessing import Process
 from time import sleep
 
@@ -29,6 +30,8 @@ _list_for_all_namespaces = {
     "secret": "list_secret_for_all_namespaces",
     "configmap": "list_config_map_for_all_namespaces"
 }
+
+_namespaces_using_file = defaultdict(list)
 
 
 def _get_file_data_and_name(full_filename, content, resource):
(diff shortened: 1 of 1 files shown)


monotek:
Sorry for the delay.
Please resolve merge conflicts.

monotek:
@pnordh 
Please resolve merge conflicts.

PR #54, opened 2019-10-26
Add capability to filter based on label value

carlosjgp:
Instead of having multiple labels configured with on multiple deployments we prefer to use the same label but with different values. Eg:

```
# Get pods with label... and any value
kubectl get configmap --selector grafana-dashboard

# Filter by label and value
kubectl get configmap --selector grafana-dashboard=team-name
```
This feature is also available on K8s Python client
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#list_namespaced_secret

As you can see the main motivation for this is to be able to use multiple sidecars on Grafana and put each team dashboard on their own folder.
We need to add more sidecars and dashboard provisioners with different folders
https://grafana.com/docs/administration/provisioning/#dashboards

I hope this makes sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - required: true
   - type: string
 
+- `LABEL_VALUE`
+  - description: The value for the label you want to filter your resources on. Don't set a value to filter by any value
+  - required: false
+  - type: string
+
 - `FOLDER`
   - description: Folder where the files should be placed
   - required: true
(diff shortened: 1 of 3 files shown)


carlosjgp:
I've tested this locally as part of the Grafana deployment with label values and the folder annotation

The old configuration still works as expected and the label value filtering also works as expected

monotek:
Please resolve merge conflicts

carlosjgp:
@monotek Conflicts resolved
And README.md file updated mentioning the new flag

PR #55, opened 2019-11-14
fix folder annotation env variable name

osela:
The name of the env variable that controls the folder annotation is `FOLDER_ANNOTATION` in the docs, but `FOLDER_ANNOTATIONS` in code.

This PR changes it to be `FOLDER_ANNOTATION` as it should according to the docs. Could change the docs instead, but the name `FOLDER_ANNOTATIONS` is confusing because there is only one annotation.

### DIFF
--- a/sidecar/sidecar.py
+++ b/sidecar/sidecar.py
@@ -8,7 +8,7 @@ from resources import listResources, watchForChanges
 def main():
     print("Starting collector")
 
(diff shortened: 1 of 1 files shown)


monotek:
I guess its better to only change the docs as this would introduce a breaking change and could giving people a hard time using this in the grafana helm chart.

PR #64, opened 2020-02-05
don't retry file when a folder can't be created due to permissions

mac9416:
1. added venv to gitignore because that's how intellij happened to set things up
1. bumped compatible k8s version in README, because things seem to work fine in 1.16
1. used os.path.join (which may be silly, this code is only going to run on linux - habits die hard)
1. handled folder-creation permissions error

Pro: no endless loop or exploding logs due to a folder that can (probably) never be created
Con: if there's some use case where write permissions are granted on the parent folder _during runtime_, k8s-sidecar won't create it until the next configmap change.

This may or may not be an awesome idea. I just found it helpful in my work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 # What?
 
-This is a docker container intended to run inside a kubernetes cluster to collect config maps with a specified label and store the included files in an local folder. It can also send an HTTP request to a specified URL after a configmap change. The main target is to be run as a sidecar container to supply an application with information from the cluster. The contained Python script is working with the Kubernetes API 1.10
+This is a docker container intended to run inside a kubernetes cluster to collect config maps with a specified label and store the included files in an local folder. It can also send an HTTP request to a specified URL after a configmap change. The main target is to be run as a sidecar container to supply an application with information from the cluster. The contained Python script is working with the Kubernetes API 1.16.
 
 # Why?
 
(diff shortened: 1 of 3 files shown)


axdotl:
Thanks for this! I think folder permissions might rarely change during runtime.

README.md
```diff
@@ -5,7 +5,7 @@
 
 # What?
 
-This is a docker container intended to run inside a kubernetes cluster to collect config maps with a specified label and store the included files in an local folder. It can also send an HTTP request to a specified URL after a configmap change. The main target is to be run as a sidecar container to supply an application with information from the cluster. The contained Python script is working with the Kubernetes API 1.10
+This is a docker container intended to run inside a kubernetes cluster to collect config maps with a specified label and store the included files in an local folder. It can also send an HTTP request to a specified URL after a configmap change. The main target is to be run as a sidecar container to supply an application with information from the cluster. The contained Python script is working with the Kubernetes API 1.16.
```
axdotl: please can you adjust the k8s version, that it become clear to users that the script will also work with lower k8s version?

monotek: Minimum version can be found here:

https://github.com/kiwigrid/k8s-sidecar/blob/master/requirements.txt

mac9416: Yep, that's much better!

mac9416: Change pushed.

PR #65, opened 2020-02-07
add log timestamps; use double-quotes for style consistency

mac9416:
I had trouble keeping track of when updates were happening, so I added timestamps to logging statements.

I also converged on double-quotes to be a bit more PEP-8-friendly. 

These are kinda preferential things. Happy to tweak!


mac9416:
@monotek is this still good to merge - just had a need for it again.  :-)

PR #71, opened 2020-04-05
Catch MaxRetryError and wait a little

uncycler:
I'm using k8s-sidecar with Istio and when the pod is spawn, it takes few seconds before it can connect to the API. If somehow the Istio sidecar is not working correctly, it generates hundred of log entries per seconds trying to connect to 172.20.0.1.

### DIFF
--- a/sidecar/resources.py
+++ b/sidecar/resources.py
@@ -147,6 +147,8 @@ def _watch_resource_loop(mode, *args):
                 listResources(*args)
                 sleep(60)
             else:
+                # Always wait 5 seconds to slow down the loop in case of exceptions
+                sleep(5)
                 _watch_resource_iterator(*args)
         except ApiException as e:
             if e.status != 500:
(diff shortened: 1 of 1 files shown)


sidecar/resources.py
```diff
@@ -154,6 +155,9 @@ def _watch_resource_loop(mode, *args):
                 raise
         except ProtocolError as e:
             print(f"{timestamp()} ProtocolError when calling kubernetes: {e}\n")
+        except MaxRetryError as e:
+            print(f"{timestamp()} MaxRetryError when calling kubernetes: {e}\n")
```
axdotl, line 159: Could you please enhance the log message to indicate that the process will wait a few seconds?

uncycler: Actually, I moved the sleep just before entering _watch_resource_iterator. This way, it catchs all exceptions.

I had the issue where the pods didn't have read secrets permission and it filled logs with millions of exceptions.

PR #77, opened 2020-05-15
Prevent from request bursts

sa-spag:
Hello there 👋

In some conditions, the sidecar may perform request bursts, especially when using the `SLEEP` mode. Therefore, I made a few tweaks:
* Avoid performing multiple successive requests for each resources
* Throttle on error when using the `SLEEP` method
* Make error throttling and `SLEEP` method sleep time configurable


monotek:
already included in: https://github.com/blablacar/k8s-sidecar/pull/1

PR #80, opened 2020-05-18
sidecar: Do not overwrite files if contents haven't changed

hwoarang:
If we update a label on the configmap/secret the sidecar will still
overwrite the file in the directory because it received an update event.
However, the contents haven't changed so we can avoid this operation.
This also has the additional benefit that other sidecars which happen to
watch the directory for changes will not trigger any reloads to the main
application.

The python implementation was inspired by [1] to ensure that we do not
chose on large reads

[1] https://www.quickprogrammingtips.com/python/how-to-calculate-sha256-hash-of-a-file-in-python.html

### DIFF
--- a/sidecar/helpers.py
+++ b/sidecar/helpers.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import hashlib
 import os
 import errno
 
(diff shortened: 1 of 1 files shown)


hwoarang:
Any thoughts on this? :)

PR #91, opened 2020-09-24
Add possibility to run SCRIPT and do it before REQ

kotchaosu:
Feature proposal. Please take a look and review. I'd like to know if you find it useful.

In current sidecar implementation I miss possibility to modify configmap exposed with sidecar via shell script, right before URL request (used it for dynamically fill piece of templated Prometheus config before reload is triggered).

By this patch one could mount script inside sidecar and specify optional "SCRIPT" field to execute it on every configmap change.


jekkel (changes requested):
Looks goods good so far, thanks for your contribution!

Two things need to be addressed though:
1. please rebase your branch on master
2. please add some kind of assertion to the test so that the build fails if the script does not get executed

sidecar/helpers.py
```diff
@@ -130,3 +131,14 @@ def unique_filename(filename, namespace, resource, resource_name):
...
+def execute(script_path):
+    try:
+        result = subprocess.run(["sh", script_path],
+                                capture_output=True,
+                                check=True)
+        print(f"Script stdout: {result.stdout}")
+        print(f"Script stderr: {result.stderr}")
+        print(f"Script exit code: {result.returncode}")
+    except subprocess.CalledProcessError as e:
+        print(f"Script failed with error: {e}")
```
jekkel, line 144:
```suggestion
        print(f"{timestamp()} Script stdout: {result.stdout}")
        print(f"{timestamp()} Script stderr: {result.stderr}")
        print(f"{timestamp()} Script exit code: {result.returncode}")
    except subprocess.CalledProcessError as e:
        print(f"{timestamp()} Script failed with error: {e}")
```

sidecar/sidecar.py
```diff
@@ -51,6 +51,7 @@ def main():
     method = os.getenv(REQ_METHOD)
     url = os.getenv(REQ_URL)
     payload = os.getenv(REQ_PAYLOAD)
+    script = os.getenv("SCRIPT")
```
jekkel, line 54: Please also use a constant for the environment variable name.

.circleci/test-in-cluster.sh
```diff
@@ -78,10 +78,12 @@ KIND_CONFIG="${CWD}/kind-config.yaml"
     kubectl cp sidecar:/tmp/hello.world /tmp/hello.world
     kubectl cp sidecar:/tmp/cm-kubelogo.png /tmp/cm-kubelogo.png
     kubectl cp sidecar:/tmp/secret-kubelogo.png /tmp/secret-kubelogo.png
+    kubectl cp sidecar:/tmp/script_result /tmp/script_result
 
     echo -n "Hello World!" | diff - /tmp/hello.world \
       && diff ${CWD}/kubelogo.png /tmp/cm-kubelogo.png \
-      && diff ${CWD}/kubelogo.png /tmp/secret-kubelogo.png
+      && diff ${CWD}/kubelogo.png /tmp/secret-kubelogo.png \
+      && diff - /tmp/script_result
```
jekkel, line 86: `-` is for stdin which is empty only by coincidence it seems we're expecting an empty file here so how about using the empty string explicitly?

jekkel (changes requested):
Looks almost good, some minor remarks left. Please also add `#minor` somewhere in a commit message as this is a new feature and the auto versioner should properly increase the version and trigger a release.

.circleci/test-in-cluster.sh
kotchaosu: my mistake, will correct that to check only if the file exists

jekkel (approved):

jekkel:
@kotchaosu I needed to create an intermediary branch to run CI but after some quick fixes to the CI job it ran successfully and I just merged it to master (#121). Version 1.11.0 should be available now. Thank you again for you contribution!

PR #104, opened 2021-02-09
Feature/adjust build and test

manu11th:
- Updating the versions of the circleci base image and docker buildx for a correct build
- Adjustments of t
- Updating the tests with newer kubernetes releases, the current version of kind and a new kind cluster config
- Making the sidecar reading binaryData from configmaps

### DIFF
--- a/sidecar/resources.py
+++ b/sidecar/resources.py
@@ -152,7 +152,7 @@ def _watch_resource_iterator(label, labelValue, targetFolder, url, method, paylo
         dataMap = event["object"].data
         if dataMap is None:
             
-            dataMap = event["object"].binaryData
+            dataMap = event["object"].binary_data
             if dataMap is None:
 
                 print(f"{timestamp()} {resource} does not have data/binaryData.")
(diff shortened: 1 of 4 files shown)


sidecar/resources.py
```diff
@@ -151,8 +151,12 @@ def _watch_resource_iterator(label, labelValue, targetFolder, url, method, paylo
         # Check if it's an empty ConfigMap or Secret
         dataMap = event["object"].data
         if dataMap is None:
-            print(f"{timestamp()} {resource} does not have data.")
-            continue
+            
+            dataMap = event["object"].binary_data
```
jekkel, line 155: This code now prefers `data` over `binary_data` and in contrast to `listResources` does not operate on both at the same time. I think we should evaluate both. Probably by looping over line 153 to 186 once for `data` and once for `binary_data` or merging the both into a single dict. I think keys present in both are conflicting anyway because are resolved to the same file name so merging might be easier.

jekkel (changes requested):
Good job! I think we have to make handling of binary and non-binary data in `_watch_resource_iterator` consistent with `listResources` but I think that's gonna be easy now.

jekkel (approved):
LGTM

PR #123, opened 2021-04-09
Set connection timeout to 60s

vsliouniaev:


jekkel:
@vsliouniaev Could you please provide a description on the PR which issues get solved with this change how the behavior is affected?

Issue #45, opened 2019-08-29
folder annotation invalid character set


r1m:
I ran into an issue trying to use the folder annotation. It raises an error saying that label value contains invalid characters.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sample-configmap
  labels:
    k8s-sidecar-target-directory: /var/lib/grafana/dashboards/test
data:
  hello.world:  Hello World!
```

```sh
The ConfigMap "sample-configmap" is invalid: metadata.labels: Invalid value: "/var/lib/grafana/dashboards/test": a valid label must be an empty string or consist of alphanumeric characters,
'-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
After digging a bit, I found that this is true if we refer to the k8s doc:
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
And it's what the k8s validation tool expects : 
https://github.com/kubernetes/apimachinery/blob/4147c925140e4e3aa51d6b46b41650821712d457/pkg/util/validation/validation.go#L88

Is it possible to find a new way of setting folder, that makes k8s validation happy ?

r1m:
I even tried to set a relative path as data keys but they suffer the same validation issue.

jekkel:
I can imagine deprecating  folder annotations in favor of a reserved data key whose content is a relative or absolute path. Whenever this data key is recognized, it's not treated as file but as configuration of the folder all other keys are created s files in.

WDYT?

cc @manu11th @pulledtim @monotek

jekkel:
@r1m In #124 @joshm91 found out that you're using it wrong. `k8s-sidecar-target-directory` should be in the *annotations* not in the labels. For annotations the restrictions you mentioned are not applied, hence path separators can be used in the value.

PR #124, opened 2021-04-21
Allow folder annotation paths to be relative or absolute

### LINKED ISSUE #101: folderAnnotation insufficient privileges
Hi guys,

I'm using grafana helm chart which comes with your image as a sidecar and having some difficulties with the folderAnnotation:

```
Error: insufficient privileges to create SRE. Skipping prometheus-overview.json.
```
here's the annotation on the configmap:
```
  annotations:                                                                                                                                                                                                                                                                      
    grafana-dashboard-folder: SRE
```
and the env var:

```
env:
- name: METHOD
- name: FOLDER_ANNOTATION                                                                                                                                                                                                                                                      
  value: grafana-dashboard-folder
- name: LABEL                                                                                                                                                                
  value: grafana_dashboard
- name: FOLDER
  value: /tmp/dashboards
- name: RESOURCE
  value: both                                                                                                                                                                                                                        image: kiwigrid/k8s-sidecar:1.1.0
```

also tried to exec into the container and create the folder manually, and it works. Not sure why the app is complaining about permissions.

```
/app $ id
uid=472 gid=472 groups=472
/app $ mkdir /tmp/dashboards/SRE ; echo $?
0
/app $
```

joshm91:
This allows the FOLDER annotation to take a relative path (relative to the default folder) which makes usage with the Grafana helm chart a lot cleaner. This would close #101

### DIFF
--- a/.circleci/test-in-cluster.sh
+++ b/.circleci/test-in-cluster.sh
@@ -88,11 +88,15 @@ KIND_CONFIG="${CWD}/kind-config.yaml"
     kubectl cp sidecar:/tmp/cm-kubelogo.png /tmp/cm-kubelogo.png
     kubectl cp sidecar:/tmp/secret-kubelogo.png /tmp/secret-kubelogo.png
     kubectl cp sidecar:/tmp/script_result /tmp/script_result
+    kubectl cp sidecar:/tmp/absolute/absolute.txt /tmp/absolute.txt
+    kubectl cp sidecar:/tmp/relative/relative.txt /tmp/relative.txt
     
     log "Verifying file content..."
     echo -n "Hello World!" | diff - /tmp/hello.world \
       && diff ${CWD}/kubelogo.png /tmp/cm-kubelogo.png \
       && diff ${CWD}/kubelogo.png /tmp/secret-kubelogo.png \
+      && echo -n "This absolutely exists" | diff - /tmp/absolute.txt \
+      && echo -n "This relatively exists" | diff - /tmp/relative.txt \
       && ls /tmp/script_result
   }
 
--- a/.circleci/test/resources.yaml
+++ b/.circleci/test/resources.yaml
@@ -29,3 +29,27 @@ data:
   script.sh: |-
      #!/bin/sh
      touch /tmp/script_result
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: absolute-configmap
+  labels:
+    findme: "yup"
+  annotations:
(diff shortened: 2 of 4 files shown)


jekkel:
Thanks for your contribution! Can you please add a corresponding test case?

joshm91:
@jekkel Thanks for the quick reply! I've added test cases to cover the original and new behaviour and updated the README.  Let me know if there's anything the needs changing.

jekkel:
I just have read into the details of #45 again. And it seems it's indeed invalid for k8s since at least 1.17 to use path separators in label values. I don't know under which circumstances or starting from which version k8s actually validates the value but I think we should not build something which knowingly violates the contract.

I wonder whether the tests would be green actually, did you run it locally? (circleCI is deactived for PRs from forks because of security considerations)

How about adding a fix for #45 as well? The possible solutions I can think of would be

1. `base64url` ([RFC 4648 §5](https://tools.ietf.org/html/rfc4648#section-5)) encoding the label value. Simple and effective.
2. moving path value to the configmap content, as proposed in #45 

WDYT?

joshm91:
Hm I did skim over #45 earlier but only just realised that they're using `k8s-sidecar-target-directory` as a _label_ and not an annotation like they should be. You're right that file separators can't be used in labels but they're [fine to use in annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) so I don't think anything is broken with the current method as suggested in #45, it looks to just be user error (unless k8s-sidecar _used_ to use labels for setting the path).

I didn't run the tests locally as I didn't have KinD set up but I did build the image and tested the functionality in our test cluster.

jekkel:
> Hm I did skim over #45 earlier but only just realised that they're using `k8s-sidecar-target-directory` as a _label_ and not an annotation like they should be. You're right that file separators can't be used in labels but they're [fine to use in annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) so I don't think anything is broken with the current method as suggested in #45, it looks to just be user error (unless k8s-sidecar _used_ to use labels for setting the path).
> 

Great catch! Thank you very much, I'll comment #45 accordingly.

jekkel (dismissed):

jekkel:
one last thing: according to CONTRIBUTING.md you should add the text `#patch` to one of your commit messages, this is going to trigger a new release with an incremented patch version then.

joshm91:
Sorry, I did read that but totally forgot to do it! I've just amended my last commit message. Thanks!

jekkel (approved):

PR #130, opened 2021-05-27
Feature flag `URL_REFRESH_INTERVAL` for update of url content #major

raghavtan:
- introduced new flag for dynamic update from *.url endpoints
- new process spawn for watching changes to *.url content dynamically
- update file on shared volume if content is updated at endpoint
- `URL_REFRESH_INTERVAL` default value is `0` - no dynamic refresh of url 
- value `>0`, the urls are dynamically updated with refresh interval of seconds `URL_REFRESH_INTERVAL` 

Use Case 
- the files created from `*.url` are currently static, in case the content at endpoint is changed then this watcher will keep the files updated in the volume
 - can be used as a dynamic backend for ambassador ingress serving custom error pages from endpoints

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - required: false
   - type: boolean
 
-- `DYNAMIC_URL`
-  - description: Set to true to enable a dynamic update of *.url content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
+- `URL_REFRESH_INTERVAL`
+  - description: Set with integer value of seconds, where 0 means `once`. set value `> 0` to enable a dynamic update with refresh interval `URL_REFRESH_INTERVAL` of *.url content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
   - required: false
-  - type: boolean
+  - type: integer
--- a/README.md
+++ b/README.md
@@ -176,3 +176,8 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - description: Set to true to enable pulling of 5XX response content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
   - required: false
   - type: boolean
+
+- `DYNAMIC_URL`
+  - description: Set to true to enable a dynamic update of *.url content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
+  - required: false
+  - type: boolean
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - type: boolean
 
 - `URL_REFRESH_INTERVAL`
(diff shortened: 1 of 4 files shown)


sidecar/sidecar.py
```diff
@@ -65,6 +66,19 @@ def main():
...
+    try:
+        if url_refresh_interval is not None:
+            url_refresh_interval = int(url_refresh_interval)
+            print(f"{timestamp()} dynamic_url content reload will be enabled. Refresh interval {url_refresh_interval}")
+        else:
+            print(f"{timestamp()} dynamic_url content reload will not be enabled.")
+            url_refresh_interval = None
+    except ValueError:
+        print(f"{timestamp()} cannot convert {URL_REFRESH_INTERVAL} to integer! Exit")
+        return -1
```
antonu17, line 80: Consider the following code

```suggestion
    try:
        url_refresh_interval = int(os.getenv(URL_REFRESH_INTERVAL, 0))
    except ValueError:
        print(f"{timestamp()} cannot convert {URL_REFRESH_INTERVAL} to integer! Exit")
        return -1

    if url_refresh_interval > 0:
        print(f"{timestamp()} dynamic_url content reload will be enabled. Refresh interval {url_refresh_interval}")
```

README.md
```diff
@@ -176,3 +176,8 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
   - description: Set to true to enable pulling of 5XX response content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
   - required: false
   - type: boolean
+
+- `URL_REFRESH_INTERVAL`
+  - description: Set with integer value of seconds, where 0 means `once`. set value `> 0` to enable a dynamic update with refresh interval `URL_REFRESH_INTERVAL` of *.url content from config map. Used in case if the filename ends with `.url` suffix (Please refer to the `*.url` feature here.)
+  - required: false
+  - type: integer
```
antonu17, line 183:
```suggestion
- `URL_REFRESH_INTERVAL`
  - description: How often to refresh the content of `.url` entries in seconds, where 0 means `once`.
  - required: false
  - default: 0
  - type: integer
```

raghavtan:
@jekkel any update on this feature, we would like to use this in our use case for serving custom error pages via ambassador

jekkel:
Hi @raghavtan , thanks for your contribution and patience.

Can't you just use `METHOD=SLEEP` ? I think this is going to do the same, isn't it? It seems this PR is duplicating some code paths unnecessarily.

Beside this I think there are some real problems:
1. I think you forgot to iterate over secrets
2. There might be some ugly race conditions because the watch process and the new url-refresh process are going to access the file system concurrently
3. if I am not mistaken after this PR one could specify `URL` for `METHOD` directly which would behave the same as `SLEEP` as the implementing code does not really restrict itself to urls.

Maybe I missed a point here so please clarify if I am mistaken.

raghavtan:
@jekkel makes sense i think `METHOD=SLEEP` is a better alternative. I think i missed out on that 
closing this PR as of now

PR #135, opened 2021-06-27
Update helpers.py

### LINKED ISSUE #76: Retries way too quickly
When you launch this sidecar next to a grafana container, you get a couple of pages of logs that the connection to k8s api is failing due to istio-proxy not having been initialised yet.

It would be great if this repo would implement an exponential decorrelated backoff to avoid this problem; wait, say 20 seconds, before crashing, too. That way, the grafana pod doesn't have to get 2 restarts before stabilising.

haf:
Fixes #76 

Docs https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html — basically, having 0.25 eats the five retries way too quickly, especially when it comes to connecting the socket over a network that might take a few seconds to establish the TCP connection.

Relevant docs:

> A backoff factor to apply between attempts after the second try (most errors are resolved immediately by a second try without a delay). urllib3 will sleep for:
> `{backoff factor} * (2 ** ({number of total retries} - 1))`
> seconds. If the backoff_factor is 0.1, then sleep() will sleep for [0.0s, 0.2s, 0.4s, …] between retries. It will never be longer than Retry.BACKOFF_MAX.
> By default, backoff is disabled (set to 0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ If the filename ends with `.url` suffix, the content will be processed as an URL
 - `REQ_RETRY_CONNECT`
   - description: How many connection-related errors to retry on
   - required: false
-  - default: 5
+  - default: 10
   - type: integer
 
 - `REQ_RETRY_READ`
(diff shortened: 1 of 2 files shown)


jekkel (changes requested):
Please adjust README.md with the changed defaults, looks good otherwise.

haf:
Feel free to make the changes you need to merge this. I’m on vacation and will not get back to this PR.

jekkel:
Nu hurry, since all the values are already configurable this change can wait a bit.

jekkel (approved):

Issue #9, opened 2018-09-06
Support for comma separated namespaces


Miouge1:
It would be nice if `NAMESPACE` would take a comma separated list of namespaces:
```
export NAMESPACE=foo,bar
```
Would scan namespace `foo` AND `bar` for ConfigMaps

Tim-Smyth:
Sorry this issue must have slipped by. Sounds like a nice feature, I currently don't have time to add it but you are welcome to do so.

jekkel:
Reopened because there's a fresh PR available.

jekkel:
released with [v1.13.0](https://github.com/kiwigrid/k8s-sidecar/releases/tag/1.13.0)

PR #136, opened 2021-07-07
Support for comma separated namespaces

Allex1:
Addresses https://github.com/kiwigrid/k8s-sidecar/issues/9

@pulledtim @jekkel @raghavtan  please review

### DIFF
--- a/sidecar/resources.py
+++ b/sidecar/resources.py
@@ -5,6 +5,7 @@ import os
 import signal
 import sys
 import traceback
+from collections import defaultdict
 from multiprocessing import Process
 from time import sleep
 
@@ -19,14 +20,13 @@ from helpers import request, write_data_to_file, remove_file, timestamp, unique_
 RESOURCE_SECRET = "secret"
 RESOURCE_CONFIGMAP = "configmap"
 
-_list_namespaced = {
+_list_namespace = defaultdict(lambda: {
     RESOURCE_SECRET: "list_namespaced_secret",
     RESOURCE_CONFIGMAP: "list_namespaced_config_map"
-}
-_list_for_all_namespaces = {
+}, {'ALL': {
     RESOURCE_SECRET: "list_secret_for_all_namespaces",
     RESOURCE_CONFIGMAP: "list_config_map_for_all_namespaces"
-}
+}})
 
 
 def signal_handler(signum, frame):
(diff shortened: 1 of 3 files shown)


sidecar/resources.py
```diff
@@ -182,14 +181,13 @@ def _update_file(data_key, data_content, dest_folder, metadata, resource,
 def _watch_resource_iterator(label, label_value, target_folder, url, method, payload,
                              current_namespace, folder_annotation, resource, unique_filenames, script, enable_5xx):
     v1 = client.CoreV1Api()
-    namespace = os.getenv("NAMESPACE", current_namespace)
     # Filter resources based on label and value or just label
     label_selector = f"{label}={label_value}" if label_value else label
 
-    if namespace == "ALL":
+    if current_namespace == "ALL":
```
jekkel, line 187: I don't think `current_namespace` can be `ALL` at this point since it is not a user provided argument but the auto detected *current* namespace.

jekkel (changes requested):
The overall approach looks good to me. But could you please unify namespace handling? Right now it's confusing where the environment variable is being evaluated and it happens in multiple places. I suggest to determine the required namespace at the very start of the sidecar.

Beside that we're dealing with 2 situations:
1. `ALL`
2. a list of specific namespaces

How about adding some object orientation to this such that there are 2 sub classes of an abstract `NamespaceConfig` class representing the 2 case. The sub classes can also provide the methods to use the k8s client for the 2 different cases so that the code making use of them doesn't have to deal with the distinction. Anyway - object oriented or not - the handling of namespaces should be simplified with the addition of the multi-namespace feature.

Last but not least please add an appropriate test case verifying the feature works as intended for all tested k8s versions.

serban21:
@jekkel I'm helping @Allex1 with this issue. Thanks for the comment.

On namespace handling. 

I agree that the namespace should be read at the start, I'll change that.

The 2 situation description is partially correct. Yes, viewed from outside it is so. However in the code I split the namespace list before starting watcher processes, so each namespace will have a process for each resource type. This is needed for processing `Watch` stream, since as far as I can tell a process is needed for each stream. Thus after starting the processes, we have the following namespace situation:

1. `ALL`
2. A single namespace

So I don't think we need to do this kind of refactor. I'll do some other changes that could improve readability. One major thing would be to see if `_watch_resource_iterator` and `list_resources` can be unified, perhaps with a similar class structure? But I'm afraid to do this now, perhaps I'll add it as a separate issue and address it afterwards.

Btw, I just realised that there is a bug on using `METHOD` `LIST`, since that is not going to start multiple processes. Fixing should be easy.

Regrading the tests, I looked at them, but I'm having problems in running them locally or on a fork CircleCI, I'm getting the same error on both: ```Must provide --username with --password-stdin```, on line ```echo "${DOCKER_PASSWORD}" | docker login --username="${DOCKER_USERNAME}" --password-stdin docker.io```. Can you tell me how to fix this, or can I get permission to push a branch into the repo, to run the tests into you CircleCI setup?

Allex1:
@jekkel can you have another look pls ?

Allex1:
@jekkel ?

jekkel (commented):
Your changes look very good, thanks for your effort! If you're not able to run the tests on your own I can do it within the org with a temporay branch...

jekkel:
@serban21 you can fix that by providing the secrets `DOCKER_USERNAME` and `DOCKER_PASSWORD` within your circleCI configuration, those will be available in the pipeline as environment variables then.

> Btw, I just realised that there is a bug on using METHOD LIST, since that is not going to start multiple processes. Fixing should be easy.

Great catch! :+1:

jekkel (approved):

PR #187, opened 2022-05-18
New config option to ignore already processed resource version.

otchilin:
`IGNORE_ALREADY_PROCESSED` option added to filter already processed resources identified by their "resourceVersion" in metadata.
This can avoid numerous checks/logs for unchanged resources. 
This feature requires kubernetes api >= v1.19.


jekkel (approved):
LGTM, thank you for this contribution! I even tend to enable this by default, wdyt?

otchilin:
> LGTM, thank you for this contribution! I even tend to enable this by default, wdyt?

Thanks for your feedback. 
I didn't enabled it by default to avoid issues for users who may expect the default behavior in their own integration. But personally I wouldn't be shocked if it was. ;)

Issue #189, opened 2022-05-24
Feature request - Ability to set log level


nhuray:
Hi,

Thanks for that project, it's really useful ! 

That project is used in the Grafana [Helm chart](https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L649) to load dashboards defined as ConfigMap into the Grafana container.

I noticed once I deployed some dashboards that the sidecar is generating a lot of logs which has an impact on cloud cost:
<img width="1709" alt="Screen Shot 2022-05-24 at 9 03 57 AM" src="https://user-images.githubusercontent.com/958983/170041577-3d562ce0-6bad-4ac4-8d60-d93d711b55d6.png">
<img width="1699" alt="Screen Shot 2022-05-24 at 9 03 36 AM" src="https://user-images.githubusercontent.com/958983/170041581-4e7fff5c-8994-4b09-b0c5-a686d34a8a46.png">

I did not find a way to control the log level passing an environment variable and looking at the code it looks like the project is not using any logging library (just using `print`). 

So would you be interested to have the ability to control the log level (I can submit a PR if needed) ?

Regards,

jekkel:
Hi @nhuray 

a contribution introducing a (light weight) logging framework and some configuration options to configure logging would be greatly appreciated.

schans:
+1 on this one. Was about to create the same ticket. Imho the default python logging framework would suffice. Might need a little bit of tuning on what level what should be logged. 

As to how to set the log level: maybe keep that in line with the other configuration and introduce a LOG_LEVEL environment variable?

schans:
Maybe something like:

`
import logging

def set_logging(l):
    loglevel = os.getenv(LOG_LEVEL, "INFO")
    numeric_level = getattr(logging, loglevel.upper(), None)
    if not isinstance(numeric_level, int):
        raise ValueError('Invalid log level: %s' % loglevel)
    logging.basicConfig(level=numeric_level, format='%(asctime)s %(levelname)s %(message)s')
`

jekkel:
released with [v1.19.0](https://github.com/kiwigrid/k8s-sidecar/releases/tag/1.19.0)

ftpd:
@nhuray @jekkel any idea, how to set it when deploying Grafana from their helm chart? It [doesn't look like](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L192-L232) I can just simply specify this env variable in my deployment.

[...]

PR #190, opened 2022-05-24
Setup logging library

nhuray:
Hi @jekkel 

This is a PR related to issue #189 to replace the logs generated using `print` by logs managed with python `logging` [library](https://docs.python.org/3/library/logging.html).

I removed all the `${timestamp()}`in the logs because the library should generate a timestamp by default.

I defined a  `LOG_LEVEL`  environment variable to control the log level globally.

I put the most verbose logs in `debug`.

Nicolas


src/helpers.py
```diff
@@ -5,11 +5,18 @@
...
 
 import requests
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
+log_level = os.getenv("LOG_LEVEL", logging.INFO)
+
+# Initialize Logger
+logger = logging.getLogger(__name__)
+logger.setLevel(log_level)
```
schans, line 18: Doesn't the setLevel method require a numeric value? 
Something like:
numeric_level = getattr(logging, log_level.upper(), None)
# ignore invalid log level setting
    if isinstance(numeric_level, int):
        logger.setLevel(numeric_level)

https://docs.python.org/3/library/logging.html#levels

nhuray: @schans Looking at the code for `setLevel` it looks `int` and `str`are supported: 

```python
def setLevel(self, level):
        """
        Set the logging level of this logger.  level must be an int or a str.
        """
        self.level = _checkLevel(level)
        self.manager._clear_cache()

def _checkLevel(level):
    if isinstance(level, int):
        rv = level
    elif str(level) == level:
        if level not in _nameToLevel:
            raise ValueError("Unknown level: %r" % level)
        rv = _nameToLevel[level]
    else:
        raise TypeError("Level not an integer or a valid string: %r" % level)
    return rv
`
``

Comment (schans) [reply src/helpers.py:18]: Today I learned ;-) Even better. We could consider some error handling though.

Comment (schans) [reply src/helpers.py:18]: May I suggest something like:

```python
# Initialize Logger
logger = logging.getLogger(__name__)

# Set log level
try:
    log_level = os.getenv("LOG_LEVEL", logging.INFO)
    log_level = log_level.upper() if isinstance(log_level, str) else log_level
    logger.setLevel(log_level)
except (ValueError, TypeError) as e:
   logger.error(f"Unable to set log level: {e}")
```

src/resources.py
```diff
@@ -17,6 +18,12 @@
 from helpers import request, write_data_to_file, remove_file, timestamp, unique_filename, CONTENT_TYPE_TEXT, \
     CONTENT_TYPE_BASE64_BINARY, execute, WATCH_SERVER_TIMEOUT, WATCH_CLIENT_TIMEOUT
 
+log_level = os.getenv("LOG_LEVEL", logging.INFO)
```
jekkel, line 21: I don't like that the environment variable is read multiple times. Can't we centralize this and provide a parsed/validated log level in a central place which gets used by all instantiated loggers then?

jekkel (changes requested):
Looks good, thanks for the contribution! I can only see a minor style issue.

src/resources.py
nhuray: We might centralize in `helpers.py` considering `resources.py`and `sidecar.py` imports `helpers.py`. 

Or we can create a dedicated `logger.py` to initialize the logger and define error handling as suggested by @schans.

Let me know what you prefer !

jekkel: I'd go for `helpers.py` :)

nhuray: Done !

src/sidecar.py
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
 
 import os
-
 from kubernetes import client, config
 from kubernetes.config.kube_config import KUBE_CONFIG_DEFAULT_LOCATION
 from requests.packages.urllib3.util.retry import Retry
 
-from helpers import timestamp, REQ_RETRY_TOTAL, REQ_RETRY_CONNECT, REQ_RETRY_READ, REQ_RETRY_BACKOFF_FACTOR
```
nhuray, line 9: I removed unused import here

src/resources.py
```diff
@@ -14,7 +14,7 @@
 from urllib3.exceptions import MaxRetryError
 from urllib3.exceptions import ProtocolError
 
-from helpers import request, write_data_to_file, remove_file, timestamp, unique_filename, CONTENT_TYPE_TEXT, \
```
nhuray, line 17: I removed unused import here

src/helpers.py
nhuray: Thanks for the suggestion @schans 👍

nhuray:
@jekkel Looking at the logs stored as Github actions [artifacts](https://github.com/kiwigrid/k8s-sidecar/suites/6651032201/artifacts/251339378), I just saw the logs does not have the format expected.

Before:
```
[2022-05-25 08:33:30] No folder annotation was provided, defaulting to k8s-sidecar-target-directory
```

After:
```
No folder annotation was provided, defaulting to k8s-sidecar-target-directory
```

It looks by default the formatter only display the [message](https://docs.python.org/3/library/logging.html#formatter-objects). So should we define a default log format for backward compatibility and define a `LOG_FORMAT` environment variable to set a format if needed ? 

Talking about backward compatibility I think we should not set the log level to `INFO` if the `LOG_LEVEL` variable is not set cause we gonna change the current behaviour (log everything).

What do you think ?

schans:
> @jekkel Looking at the logs stored as Github actions [artifacts](https://github.com/kiwigrid/k8s-sidecar/suites/6651032201/artifacts/251339378), I just saw the logs does not have the format expected.
> 
> Before:
> 
> ```
> [2022-05-25 08:33:30] No folder annotation was provided, defaulting to k8s-sidecar-target-directory
> ```
> 
> After:
> 
> ```
> No folder annotation was provided, defaulting to k8s-sidecar-target-directory
> ```
> 
> It looks by default the formatter only display the [message](https://docs.python.org/3/library/logging.html#formatter-objects). So should we define a default log format for backward compatibility and define a `LOG_FORMAT` environment variable to set a format if needed ?

Something this might do the trick:
```python
logging.basicConfig(format='%(asctime)s %(levelname)s %(message)s')
```
We could make it configurable. For us that wouldn't bring that much extra value. Time and level are useful. 

We would prefer logging in json format following ECS standard, but that's an easy improvement after this PR is in.

 > Talking about backward compatibility I think we should not set the log level to `INFO` if the `LOG_LEVEL` variable is not set cause we gonna change the current behaviour (log everything).
> 
> What do you think ?

INFO or WARING default level makes sense to me. Log everything (debug level) is a bit too much and to some extend the direct reason for this PR ;-) But since it's configurable it wouldn't make that much of a difference.

src/helpers.py
```diff
@@ -35,6 +36,29 @@
...
+    """
+    log = logging.getLogger(name)
+    try:
+        log_level = os.getenv("LOG_LEVEL", logging.INFO)
+        log_level = log_level.upper() if isinstance(log_level, str) else log_level
+        log.setLevel(log_level)
+        return log
+    except (ValueError, TypeError) as e:
+        log.error(f"Unable to set log level: {e}")
+        exit(1)
```
schans, line 55: I would prefer not to exit here. It just falls back to INFO (default) level and can run just fine.

schans:
Or if we want to have timestamps in RFC3339:

```python
logging.basicConfig(format="%(asctime)s %(levelname)s %(message)s", datefmt='%Y-%m-%dT%H:%M:%S%z')
```

That would result in:
```
>>> logging.error('this is an error log message')
2022-05-25T14:17:20+0200 ERROR this is an error log message
```

src/helpers.py
nhuray: make sense. In this case I would define the log as a warning:

```
log.warning("Unable to set log level: {e}. Define default level to INFO")
```

nhuray:
@schans @jekkel 

Finally I added a `logger.py`  in order to centralize the logging capabilities.

I implemented 2 formatters which I guess should be good enough for now:
 - `LOGFMT`: Format log using [logfmt](https://brandur.org/logfmt) format
 - `JSON`: Format log in json

This is an example of the log generated:

**Logfmt**
```
time=2022-05-25T23:15:45.581453 level=WARNING msg="No folder annotation was provided, defaulting to k8s-sidecar-target-directory"
```

**Json**
```
{"time": "2022-05-25T23:24:20.836608", "level": "WARNING", "msg": "No folder annotation was provided, defaulting to k8s-sidecar-target-directory"}
```

@schans I did not implemented the ECS format cause it looks quite verbose to me for a so small project.

src/logger.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Instantiate a logger with the specified name, level and formatter.
+
+    :param name: logger name
+    :param level: logger level (default INFO)
+    :param fmt: logger formatter (default LOGFMT)
+    :return: a logger instance
+    """
+    log = logging.getLogger(name)
+    ch = logging.StreamHandler()
```
jekkel, line 94: I think configuring the root logger globally instead of each created loggers individually might improve readability a bit, wdyt?

src/logger.py
```diff
@@ -0,0 +1,108 @@
...
+        log_format_msg = list(
+            map(lambda key:
+                '%s="%s"' % (key, fields.get(key).replace('"', '\\"').replace('\n', '\\n'))
+                if key in ('msg', 'exception', 'stack')
+                else '%s=%s' % (key, fields.get(key)), fields.keys())
+        )
+        return " ".join(log_format_msg)
+
+
+class JsonFormatter(BaseFormatter):
```
jekkel, line 62: Shouldn't we just use `python-json-logger` (https://pypi.org/project/python-json-logger/) for JSON logging instead of implementing it here?

src/logger.py
```diff
@@ -0,0 +1,108 @@
...
+        fields['level'] = record.levelname
+        fields['msg'] = record.getMessage()
+        if record.exc_info:
+            fields['exception'] = self.formatException(record.exc_info)
+        if record.stack_info:
+            fields['stack'] = self.formatStack(record.exc_info)
+        return fields
+
+
+class LogfmtFormatter(BaseFormatter):
```
jekkel, line 44: I think configuring the [default log formatter](https://docs.python.org/3/library/logging.html#formatter-objects) with suitable format strings should be good enough. I don't see why we need a custom formatter.

README.md
```diff
@@ -81,3 +81,5 @@ If the filename ends with `.url` suffix, the content will be processed as a URL
...
+| `LOG_LEVEL`                | Set the logging level. (DEBUG, INFO, WARN, ERROR, CRITICAL)                                                                                                                                                                                                                                                                         | false    | `INFO`                                    | string  |
+| `LOG_FORMAT`               | Set a log format. (JSON or LOGFMT)                                                                                                                                                                                                                                                                                                  | false    | `LOGFMT`                                  | string  |
```
jekkel, line 85: How about making this config then " ... `JSON` or a format string for the [default formatter](https://docs.python.org/3/library/logging.html#logging.Formatter)" ?

src/helpers.py
```diff
@@ -65,7 +70,7 @@ def write_data_to_file(folder, filename, data, data_type=CONTENT_TYPE_TEXT):
                 sha256_hash_cur.update(byte_block)
 
         if sha256_hash_new.hexdigest() == sha256_hash_cur.hexdigest():
-            print(f"{timestamp()} Contents of {filename} haven't changed. Not overwriting existing file")
+            logger.warning(f"Contents of {filename} haven't changed. Not overwriting existing file")
```
jekkel, line 73: -> debug level, updates of config/map secrets not touching all entries can expected and should not create `WARNING` logs.

jekkel (changes requested):

src/resources.py
```diff
@@ -92,12 +96,12 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
         # Avoid numerous logs about useless resource processing each time the LIST loop reconnects
         if ignore_already_processed:
             if _resources_version_map.get(metadata.namespace + metadata.name) == metadata.resource_version:
-                # print(f"{timestamp()} Ignoring {resource} {metadata.namespace}/{metadata.name}")
+                # logger.debug(f"{timestamp()} Ignoring {resource} {metadata.namespace}/{metadata.name}")
```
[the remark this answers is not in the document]
nhuray: good point

src/logger.py
nhuray: Might be a good option as well. I was just reticent to add dependencies to the project to keep it lean.

But we might consider those library are sufficiently small to not harm ! 

So up to you @jekkel ! Just le me know what you prefer

src/logger.py
nhuray: Here is a custom formatter to use LogFmt fomat (key and value). 

We can also use a format string like `time=%(asctime)s level=%(levelname)s msg=%(message)s` .

I centralize the log fields  in `BaseFormatter` in order to use the same for JsonFormatter and LogFmtFormatter but I can also revisit that to simplify things.

README.md
nhuray: You mean make Json the default format ? 

Absolutely, I personally like LogFmt because it's more human readable and les verbose that Json. But Json is probably more versatile for parsing by log aggregator (Loki, ELK ...)

src/helpers.py
nhuray: agree

src/logger.py
nhuray: agree

src/logger.py
```diff
@@ -0,0 +1,108 @@
...
+         - time: ISO 8601 format with milliseconds (from datetime.isoFormat)
+         - level: Log level in uppercase letter
+         - msg: the log message
+         - exception: exception if the logger is specifying exc_info=True
+         - stack: stack information if the logger is specifying stack_info=True
+        """
+        fields = dict()
+        fields['time'] = self.formatTime(record)
+        fields['level'] = record.levelname
+        fields['msg'] = record.getMessage()
```
schans, line 36: We could chose to align these fields with the ECS standard. That would entail:
time -> @timestamp
level -> log.level
msg -> message

For an example see:
https://www.elastic.co/guide/en/ecs-logging/python/current/installation.html
Full ref:
https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html

src/logger.py
```diff
@@ -0,0 +1,108 @@
...
+    def formatTime(self, record):
+        """
+        Return the creation time of the specified LogRecord as formatted text.
+        This method should be called from format() by a formatter which
+        wants to make use of a formatted time.
+        It provides an ISO 8601 format with milliseconds.
+        """
+        ct = self.converter(record.created)
+        ct_with_ms = time.mktime(ct) + (record.msecs / 1000)
+        ct_full = datetime.datetime.fromtimestamp(ct_with_ms)
```
schans: Could we add the timezone info here? That would reduce the confusion when trying to correlate different log events. 
 Is that info available in the record? Or is the timestamp always in the system timezone? In that case we could use something link:

```python
ct_full = datetime.datetime.fromtimestamp(ct_with_ms, datetime.datetime.utcnow().astimezone().tzinfo)
```

src/logger.py
nhuray: I just configured a single logger named `k8s-sidecar` that I'm using everywhere.

src/logger.py
nhuray: I added the dependency you suggested to log in Json, along another one to format in LogFmt. 

If we don't need the second one let me know I can remove it !

README.md
nhuray: I changed to have `JSON` has the default

src/logger.py
nhuray: @schans I understand the need to have a uniform format but I'm not sure ECS is widely used in community. 

So why use more ECS over [Gelf](https://docs.graylog.org/docs/gelf) or [Json Well Format](https://github.com/JSONWellLogFormat/JSONWellLogFormat) ?

From my perspective what's is important is to have a Json format cause all log collectors / aggregators supports JSON. 

I implemented LogFmt as well, cause it's a format widely used in Go community, but I let the maintainers deciding to remove it or not !

src/logger.py
nhuray: @schans I just made a change to expose a `LOG_TZ` variable to choose time formatted in local timezone or UTC timezone. 

The `time` fields is formatted in ISO 8601 format (e.g `2022-05-28T14:24:04.037058+00:00`)

src/sidecar.py
```diff
@@ -87,17 +91,17 @@ def main():
             v_minor = re.sub(r'\D', '', version.minor)
 
             if len(v_major) and len(v_minor) and (int(v_major) > 1 or (int(v_major) == 1 and int(v_minor) >= 19)):
-                print(f"{timestamp()} Ignore already processed resource version will be enabled.")
+                logger.debug("Ignore already processed resource version will be enabled.")
```
otchilin, line 94: Hi !
Perhaps you may log these ones in 'info' level and 'warning' for the api compatibility issue instead of debug.

Thanks.
Regards.

src/logger.py
schans: Thanks.

From my local tests and what I understand from the docs, the isoformat() function will only add the TZ info if that is present in the timestamp (the +00:00 part), but maybe that's something local to my setup.

src/logger.py
schans: I guess it's a matter of perspective on what you are used to ;-)

Looking at the Gelf format I wouldn't be a fan of numeric log levels and unix timestamps. 

For the Json Well Format, I'm not sure if that's the one you wanted to link to. It seems specific to (drilling) well measurements in the oil and gas industry.

I don't mind that much about the field names. I think most log aggregation software out there can easily rename and transform. Imho timestamp is a bit more specific than time. If we use msg instead of message you could also opt for ts instead of timestamp. I have seen that plenty as well. But if you prefer to keep the names as they are now that is perfectly fine as well.

nhuray: My reasoning about using `time`, `level` and `msg` is to have fields semantically comprehensible for humans and sufficiently small to limit the size it takes to store the fields in a reversed index (eg. Elasticsearch, Loki...).

That format is what used the golang lib [Logrus](https://github.com/sirupsen/logrus). I saw a lot of Kubernetes projects using `logrus` for logging so considering that project is also for running in Kubernetes I adopt the same log format.

src/logger.py
nhuray: I'm using the record `created` (POSIX timestamp) and format it like that:
```
datetime.fromtimestamp(record.created, log_tz).isoformat()
```
So I format to IS0 8601 format a `datetime` with `tzinfo`.

From my tests it gives:

**UTC TZ**
```
{"time": "2022-05-29T11:22:11.377220+00:00", "level": "INFO", "msg": "Starting collector"}
```


**LOCAL TZ**
```
{"time": "2022-05-29T07:21:56.590056-04:00", "level": "INFO", "msg": "Starting collector"}
```

src/sidecar.py
nhuray: Good catch ! Thanks

src/logger.py
schans: Nice! Perfect.

src/logger.py
schans: Aha. Yes, that makes sense.

nhuray:
@jekkel Is there something blocking to merge that PR now ?

nhuray:
> To me everything looks good to go.
> 
> Thanks @nhuray for the work and patience dealing with all my comments!

Thanks for you review

jekkel (approved):

Issue #188, opened 2022-05-24
`IGNORE_ALREADY_PROCESSED` fails on EKS 1.21


ccodreanu:
Hi!

I wanted to report an issue with the 1.18.0 version and specifically with the flag in title.

When running on EKS 1.21 the container errors out due to:

```
Traceback (most recent call last):
  File "/app/sidecar.py", line 142, in <module>
    main()
  File "/app/sidecar.py", line 82, in main
    if int(version.major) > 1 or (int(version.major) == 1 and int(version.minor) >= 19):
ValueError: invalid literal for int() with base 10: '21+'
```

Thanks!

jekkel:
@ccodreanu Can you please post the output of
```
kubectl version --output=yaml
```

@otchilin You probably want to have look into this?

ccodreanu:
> @ccodreanu Can you please post the output of
> 
> ```
> kubectl version --output=yaml
> ```

```
clientVersion:
  buildDate: "2021-11-17T15:48:33Z"
  compiler: gc
  gitCommit: b695d79d4f967c403a96986f1750a35eb75e75f1
  gitTreeState: clean
  gitVersion: v1.22.4
  goVersion: go1.16.10
  major: "1"
  minor: "22"
  platform: darwin/amd64
serverVersion:
  buildDate: "2022-02-17T16:36:28Z"
  compiler: gc
  gitCommit: eb09fc479c1b2bfcc35c47416efb36f1b9052d58
  gitTreeState: clean
  gitVersion: v1.21.9-eks-0d102a7
  goVersion: go1.16.12
  major: "1"
  minor: 21+
  platform: linux/amd64

jekkel:
@ccodreanu please test with [v1.18.1](https://github.com/kiwigrid/k8s-sidecar/releases/tag/1.18.1)

PR #191, opened 2022-05-24
Bug #188 fix. Kubernetes api version detection improved.

otchilin:
K8S API version detection reworked to be safer and compatible with non digit major/minor version content.
Only numbers are retained and empty version are rejected.

### DIFF
--- a/src/sidecar.py
+++ b/src/sidecar.py
@@ -3,11 +3,13 @@
 import os
 
 from kubernetes import client, config
+from kubernetes.client import ApiException
 from kubernetes.config.kube_config import KUBE_CONFIG_DEFAULT_LOCATION
 from requests.packages.urllib3.util.retry import Retry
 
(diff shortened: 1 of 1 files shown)


src/sidecar.py
```diff
@@ -75,21 +77,28 @@ def main():
...
+            if len(v_major) and len(v_minor) and (int(v_major) > 1 or (int(v_major) == 1 and int(v_minor) >= 19)):
+                print(f"{timestamp()} Ignore already processed resource version will be enabled.")
+                ignore_already_processed = True
+            else:
+                print(
+                    f"{timestamp()} Can't enable 'ignore already processed resource version', "
+                    f"kubernetes api version (%s) is lower than v1.19 or unrecognized format." % version.git_version)
+
+        except ApiException as e:
+            print("Exception when calling VersionApi->get_code: %s\n" % e)
```
jekkel, line 98: please add the timestamp to this log message.

jekkel (changes requested):

jekkel (approved):

PR #192, opened 2022-06-07
enhancement: Add token based auth as alternative to basic-auth for http requests.

jansalapatek:
Added two new variables to manage token based authentication which for example is required for using k8s-sidecar with git(lab), since it does not support basic-auth.


jansalapatek:
I cannot label the PR, but it is an enhancement.

README.md
```diff
@@ -69,9 +69,10 @@ If the filename ends with `.url` suffix, the content will be processed as a URL
...
+| `REQ_PASSWORD`             | Password to use for basic authentication for requests to `REQ_URL` and for `*.url` triggered requests                                                                                                                                                                                                                               | false    | -                                         | string  || `SCRIPT`                   | Absolute path to shell script to execute after a configmap got reloaded. It runs before calls to `REQ_URI`                                                                                                                                                                                                                          | false    | -                                         | string  |
```
jekkel, line 75: you removed a line break which should be there.

README.md
```diff
@@ -47,7 +47,7 @@ metadata:
     k8s-sidecar-target-directory: "/path/to/target/directory"
 ```
 
-If the filename ends with `.url` suffix, the content will be processed as a URL which the target file contents will be downloaded from.
+If the filename ends with `.url` suffix, the content will be processed as a URL which the target file contents will be downloaded from. If the source requires authentication, currently http basic-auth and header based token authentication are supported. Please see the variables below for details.
```
jekkel, line 50: here it's described as header based token auth but further down it's described as parameters and if I read the impl correctly it's actually doing query parameters, right?

jekkel (changes requested):

jekkel:
Would you be able to add a test for your scenario?

README.md
jansalapatek: Fixed accordingly.

README.md
jansalapatek: Fixed.

src/helpers.py
```diff
@@ -102,10 +102,33 @@ def request(url, method, enable_5xx=False, payload=None):
...
+        logger.debug(f"Basic-auth configured.  password: {password}.")
     else:
+        logger.info(f"No authorization tokens set (no basic-auth and no private token).")
         auth = None
+        params = None
+
+    if params:
+      logger.debug(f"Request params are set as follows: {params}")
+    else:
+      logger.info(f"No request params are set.")
```
jekkel, line 131: remove. Or lower to `debug`.

src/helpers.py
```diff
@@ -102,10 +102,33 @@ def request(url, method, enable_5xx=False, payload=None):
...
+
+    if req_token_key and req_token_value:
+        auth = None
+        params[ str(req_token_key) ] = req_token_value
+        logger.info(f"Token based authorization configured. Token key: {req_token_key}.")
+        logger.debug(f"Token value {params[str(req_token_key)]}.")
+    elif username and password:
         auth = (username, password)
+        params = None
+        logger.info(f"Basic-auth configured.  username: {username}.")
```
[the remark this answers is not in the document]
jansalapatek: lowered to debug

src/helpers.py
```diff
@@ -102,10 +102,33 @@ def request(url, method, enable_5xx=False, payload=None):
...
+
+    if req_token_value and not req_token_key:
+      req_token_key = "private_token"
+      logger.info(f"Request token value is set, but key is not. Setting request token key to default value {req_token_key}.")
+
+    if req_token_key and req_token_value:
+        auth = None
+        params[ str(req_token_key) ] = req_token_value
+        logger.info(f"Token based authorization configured. Token key: {req_token_key}.")
+        logger.debug(f"Token value {params[str(req_token_key)]}.")
```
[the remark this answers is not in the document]
jansalapatek: removed

jekkel:
There's no real test concept. What we have right now is a e2e test using Kind and a longish bash script whose exit code is relevant for fail or pass. Adding a test case would mean to adapt the script accordingly or introduce a sound test concept. I'd be ok with either.

jansalapatek:
I do not exactly know how to test it locally, but I think I got the general concept.

test/resources/sidecar.yaml
```diff
@@ -105,6 +105,51 @@ spec:
...
+  namespace: default
+spec:
+  serviceAccountName: sample-acc
+  containers:
+  - name: sidecar
+    image: kiwigrid/k8s-sidecar:testing
+    volumeMounts:
+    - name: shared-volume
+      mountPath: /tmp/
+    - name: script-volume
```
jekkel, line 129: I think we don't need the script here.

test/resources/sidecar.yaml
```diff
@@ -105,6 +105,51 @@ spec:
...
+    - secretRef:
+        name: token-auth-sample-secret      
+    env:
+    - name: LABEL
+      value: "findme"
+    - name: FOLDER
+      value: /tmp/
+    - name: RESOURCE
+      value: both
+    - name: SCRIPT
```
jekkel, line 142: and here

test/resources/sidecar.yaml
```diff
@@ -105,6 +105,51 @@ spec:
...
+      value: "/opt/script.sh"
+  volumes:
+  - name: shared-volume
+    emptyDir: {}
+  - name: script-volume
```
jekkel, line 147: and here

test/server/server.py
```diff
@@ -27,3 +38,9 @@ async def read_item():
 @app.post("/503", status_code=503)
 async def read_item():
     return 503
+
+
+@app.get("/200/api_key")
```
jekkel, line 43: could you add another endpoint with a different API key so we can add a negative test case as well? That would be awesome!

jekkel (approved):
Great job, thank you for your contribution! Approving already but do you think a negative test for wrong api key would be worth the effort?

jekkel:
@jansalapatek it seems the test breaks: https://github.com/kiwigrid/k8s-sidecar/runs/6875936389?check_suite_focus=true

test/resources/sidecar.yaml
jansalapatek: good point. removed the script

test/resources/sidecar.yaml
jansalapatek: good point. removed the script

test/resources/sidecar.yaml
jansalapatek: good point. removed the script

test/resources/sidecar.yaml
jansalapatek: also switched resource to configmap

test/server/server.py
jansalapatek: added a new container with wrong credentials/secret and used the same endpoint

jekkel:
Not yet: https://github.com/kiwigrid/k8s-sidecar/runs/6878420651?check_suite_focus=true

test/server/server.py
```diff
@@ -1,8 +1,19 @@
-from fastapi import FastAPI
-import uvicorn
+#from http.client import HTTPException, HTTPResponse
+from fastapi import FastAPI, Security, Depends, HTTPException
+from fastapi.security.api_key import APIKeyQuery, APIKey
+#import uvicorn
```
jekkel, line 4: remove unused imports

jekkel (approved):

jekkel (approved):

jekkel:
I think we only need to get the build green before merging/releasing.

jekkel (approved):

jansalapatek:
Build failed, because python image could not be fetched from registry :-/ server had a 503.

jekkel:
> Build failed, because python image could not be fetched from registry :-/ server had a 503.

New build running...

PR #217, opened 2022-09-27
Introduce ability to ignore requests on initial "ADDED" events

### LINKED ISSUE #216: Ability to skip HTTP requests on initial "ADDED" events
Hi,

Awesome project! This is exactly what I was looking for.

I use the request functionality to be notifier every time one of the watched ConfigMaps or Secrets is updated. However, during the container start, I'm getting multiple requests to my service about "ADDED" event, which I would like to avoid. It would be great to have it configurable so that my service is notified only when any other event occurs except the initial "Added" one.

I can contribute the feature as I already updated the code on my fork. Stay tuned!

pkosiec:
Resolves #216.

Let me know what you think!


src/resources.py
```diff
@@ -114,6 +117,10 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
     if script and files_changed:
         execute(script)
 
+    if request_ignore_initial_event and ignore_request:
```
jekkel, line 120: This gets executed after processing all the items in the batch. Only if a single one hasn't been seen before, we skip executing the request, despite other already known items might have received updates.

The issue here is that we're doing the request only once for a batch whenever at least a single item changed. Following this logic, `request_ignore_initial_event` should only count as flag to determining whether something has changed, i.e. only when *all* items in the batch receive their initial version or an insignificant update (no changes to content), the request gets skipped.

src/resources.py
```diff
@@ -114,6 +117,10 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
...
+        logger.debug(f"Ignoring sending request for initial list {resource} {metadata.namespace}/{metadata.name}")
```
jekkel, line 121: At this point in the code `metadata` points to the last item of the loop. Which might be a different one than the item which made `ignore_request` flip to `True`.

src/resources.py
```diff
@@ -257,6 +269,10 @@ def _watch_resource_iterator(label, label_value, target_folder, request_url, req
         if script and files_changed:
             execute(script)
 
+        if request_ignore_initial_event and ignore_request:
```
jekkel, line 272: same story here.

README.md
```diff
@@ -68,6 +68,7 @@ If the filename ends with `.url` suffix, the content will be processed as a URL
...
 | `REQ_RETRY_BACKOFF_FACTOR` | A backoff factor to apply between attempts after the second try for any http request (`.url` triggered requests, requests to `REQ_URI` and k8s api requests)                                                                                                                                                                        | false    | `1.1`                                     | float   |
+| `REQ_IGNORE_INITIAL_EVENT` | Set to `true` to ignore requests for first events or initial list for configmaps and secrets events. Applicable only when `IGNORE_ALREADY_PROCESSED` is enabled.                                                                                                                                                                    | false    | `false`                                   | boolean |
```
jekkel, line 71: I think we should explain better under which circumstances a request will be skipped (accounting in the batch logic).

jekkel (changes requested):
I think this still needs some small changes.

Furthermore, can you potentially write a small test scenario? This would involve:
* adjusting `./test/server/server.py` to be able to identify whether a call has been made
* adjust `./test/resources/sidecar.yaml` to configure a new sidecar instance or reconfigure an existing one to skip the request
* in the [github test workflow](https://github.com/kiwigrid/k8s-sidecar/blob/master/.github/workflows/build_and_test.yaml#L90-L129) add appropriate file downloads and verification.

pkosiec:
Hi @jekkel, thank you for review! I applied changes to the app and tested that locally. Now it should work as expected for the batched items as well.

When it comes to the integration tests, would you be able to help me with that? Unfortunately, I'm not fluent in Python and that would be difficult for me. Thank you in advance 🙇

README.md
pkosiec: Improved - please take a look now 🙂

src/resources.py
pkosiec: Thanks, you're correct. I inverted the bool to make sure it works properly for batching items.

pkosiec:
Hey @jekkel, could you please take a look at this PR once again? Thank you!

jekkel:
> When it comes to the integration tests, would you be able to help me with that? Unfortunately, I'm not fluent in Python and that would be difficult for me. Thank you in advance

Most of the test logic is bash anyway.

Issue #230, opened 2022-11-16
Files are not deleted when event_type is MODIFIED and k8s-sidecar-target-directory has changed


AvihaiSam:
When changing configmap's / secret's `k8s-sidecar-target-directory` annotation, the original file is not deleted.
Instead - there are 2 identical files in different locations.

jekkel:
fixed with #231

PR #231, opened 2022-11-16
support dest_folder state

AvihaiSam:
When the only thing that changes in configmap/secret is `k8s-sidecar-target-directory` annotation - the old file doesn't get deleted.

Added dest_folder mechanism to handle deletion of the described above.
Issue #230

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -120,7 +120,7 @@ jobs:
           kubectl cp sidecar:/tmp/script_result /tmp/script_result
           kubectl cp sidecar:/tmp/absolute/absolute.txt /tmp/absolute.txt
           kubectl cp sidecar:/tmp/relative/relative.txt /tmp/relative.txt
-          kubectl cp sidecar:/tmp/old-dir/change-dir.txt /tmp/change-dir.txt
+          kubectl cp sidecar:/tmp/orig-dir/change-dir.txt /tmp/change-dir.txt
           kubectl cp sidecar:/tmp/500.txt /tmp/500.txt || true
           kubectl cp sidecar:/tmp/secured.txt /tmp/secured.txt
 
@@ -133,7 +133,7 @@ jobs:
           # absolute path in configmap points to /tmp in 'absolute-configmap'
           kubectl cp sidecar-5xx:/tmp/absolute/absolute.txt /tmp/5xx/absolute.txt
           kubectl cp sidecar-5xx:/tmp-5xx/relative/relative.txt /tmp/5xx/relative.txt
-          kubectl cp sidecar-5xx:/tmp-5xx/old-dir/change-dir.txt /tmp/5xx/change-dir.txt
+          kubectl cp sidecar-5xx:/tmp-5xx/orig-dir/change-dir.txt /tmp/5xx/change-dir.txt
           kubectl cp sidecar-5xx:/tmp-5xx/500.txt /tmp/5xx/500.txt
           kubectl cp sidecar-5xx:/tmp-5xx/secured.txt /tmp/5xx/secured.txt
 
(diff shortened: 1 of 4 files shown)


AvihaiSam:
#230

jekkel:
Thanks for thecontribution! Could you please add a corresponding test?

AvihaiSam:
> Thanks for the contribution! Could you please add a corresponding test?

@jekkel i have added them months ago.... can you please?

.github/workflows/build_and_test.yaml
```diff
@@ -172,11 +175,14 @@ jobs:
...
+            echo -n "This new-dir exists" | diff - /tmp/5xx/change-dir.txt &&
+            echo -n "500" | diff - /tmp/5xx/500.txt &&
             ls /tmp/5xx/script_result &&
             kubectl exec sidecar -- sh -c "ls /tmp/" &&
             kubectl exec sidecar -- sh -c "! test -e /tmp/hello.world" && kubectl exec sidecar -- sh -c "test -e /tmp/change-hello.world" &&
             kubectl exec sidecar -- sh -c "! test -e /tmp/cm-kubelogo.png" && kubectl exec sidecar -- sh -c "test -e  /tmp/change-cm-kubelogo.png" &&
             kubectl exec sidecar -- sh -c "! test -e /tmp/secret-kubelogo.png" && kubectl exec sidecar -- sh -c "test -e /tmp/change-secret-kubelogo.png" &&
             kubectl exec sidecar -- sh -c "! test -e /tmp/absolute/absolute.txt" && kubectl exec sidecar -- sh -c "test -e /tmp/absolute/change-absolute.txt" &&
             kubectl exec sidecar -- sh -c "! test -e /tmp/relative/relative.txt" && kubectl exec sidecar -- sh -c "test -e /tmp/relative/change-relative.txt"
+            kubectl exec sidecar -- sh -c "! test -e /tmp/relative/relative.txt" && kubectl exec sidecar -- sh -c "test -e /tmp/relative/change-relative.txt"
```
jekkel, line 187: you need to concatenate all assertions with `&&` into a huge statement (or refactor to use some better assertion mechanisms...) otherwise failures in previous lines won't be relevant anymore to the step result.

src/resources.py
```diff
@@ -193,10 +199,11 @@ def _process_config_map(dest_folder, config_map, resource, unique_filenames, ena
             is_removed)
     if old_config_map.data is not None and not is_removed:
         for key in set(old_config_map.data.keys()) & set(config_map.data or {}):
-            old_config_map.data.pop(key)
+            if old_dest_folder == dest_folder:
```
jekkel, line 202: We leave only those keys (files) for removal which are not in the new resource anymore, entries not mentioned shall be removed. Your change disables this if the folder changed. So this means old files won't be deleted on destination folder changes. This seems to be an undesired behavior. I'd expect *all* files of the old configmap to be deleted if the destination folder changed, don't you agree?

jekkel (changes requested):
Sorry for the delay, I reviewed the changes and made some comments. Will also try to re-run CI.

src/resources.py
AvihaiSam: I'm not really sure what you mean.
Each key is popped out if it doesn't need to change (remove/create).
This condition makes sure that a key is a candidate for deletion only if the folder has changed! (it is not `is_removed`)
On the other hand - if the folder has changed -> you need to remove the original file (from old folder).

.github/workflows/build_and_test.yaml
AvihaiSam: i have fixed the tests now. should be OK.

src/resources.py
jekkel: `is_removed` relates to the full resource (either configmap or secret), not the individual keys.
So the whole branch is only reached if the resource has been changed or added.
To clean-up files which we created initially which are not valid anymore, we iterate on the intersection of the resource keys (the files which are not to be removed) and remove those keys from `old_config_map.data` and then iterate on the old map (with reduced keyset representing only the files which are to be removed) as a deletion (hard coded `True` for `is_deleted`).

The point is that if the destination folder changed, **all** "old" files need to be removed, not only the intersection of keys.

But maybe I am mixing things up here, the code grew quite a bit and would benefit from a refactoring. Your assertions should cover this case.

.github/workflows/build_and_test.yaml
jekkel: Unfortunately this now fails when trying to deploy the kind k8s cluster (I suspect a github change on the build agents). I'll try to fix this in another branch...

src/resources.py
AvihaiSam: You are correct, if the destination has changed - non of the keys will be popped out => therefor will be deleted.
The only thing I'd change is move the `key` loop inside the folder destination condition.
Refactoring sounds reasonable...

.github/workflows/build_and_test.yaml
jekkel: please rebase on master, then CI should be ok again

src/resources.py
jekkel: Yes please, CI now also succeeded.

AvihaiSam: Done.

jekkel (approved):

PR #232, opened 2022-11-18
Fix CVE-2022-43680

NicolasStr:
Fixing expat v2.4.9-r0 to latest by adding an `apk add --upgrade expat` after installing python requirements.

fixes CVE-2022-43680: https://avd.aquasec.com/nvd/2022/cve-2022-43680/


jekkel:
Is upstream affected as well? Can we rebuild on top of a fixed base image?

Context: I don't want to clutter the docker file with CVE induced upgrades as it's an issue of the base image and fixing it in on our side increases the image size every time (since the vulnerable version is only shadowed, not removed).

NicolasStr:
It looks like an issue with python packages being installed. Alpine image does not have that CVE.

jekkel:
@NicolasStr  is the latest version still affected?

jekkel:
Fixed since `v1.21.1`.

Issue #250, opened 2023-01-19
Cannot download binary files


svolland-csgroup:
There are some issues that prevent downloading binary files, i.e. using `binaryData` config maps with `*.url` keys.
- download url is not correctly decoded (base64 binary data is not decoded to a string)
- data is downloaded as text, which fails when writing the file in binary mode

The issue can be reproduced with the following config map to download kubectl (just as an example)

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: download-kubectl
  labels:
     download_binary: "1"
binaryData:
  # Download https://storage.googleapis.com/kubernetes-release/release/v1.26.0/bin/linux/amd64/kubectl
  kubectl.url: "aHR0cHM6Ly9zdG9yYWdlLmdvb2dsZWFwaXMuY29tL2t1YmVybmV0ZXMtcmVsZWFzZS9yZWxlYXNlL3YxLjI2LjAvYmluL2xpbnV4L2FtZDY0L2t1YmVjdGw="
```

jekkel:
closed by #251

PR #251, opened 2023-01-19
Fix handling of binaryData .url keys

svolland-csgroup:
This fixes issue #250

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -6,6 +6,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3          
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Build sidecar
@@ -20,7 +22,7 @@ jobs:
       - name: Build dummy server
         uses: docker/build-push-action@v3
         with:
-          context: "{{defaultContext}}:test/server"
+          context: "test/server"
           push: false
           outputs: type=docker,dest=/tmp/dummy-server.tar
           tags: "dummy-server:1.0.0"
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -14,6 +14,9 @@ jobs:
           push: false
           outputs: type=docker,dest=/tmp/k8s-sidecar.tar
           tags: "kiwigrid/k8s-sidecar:testing"
+      - name: Prepare dummy server static resources
+        run: |
+          cp test/kubelogo.png test/server/static/
       - name: Build dummy server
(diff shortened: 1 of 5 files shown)


jekkel (commented):
Thanks for your contribution! The fixes look good. Could you please add reproducers for both issues to the test in the github workflow?

svolland-csgroup:
Tests added, in fact it is only one issue (by fixing only the first point you hit the second point. Both are fixed)

test/resources/resources.yaml
```diff
@@ -72,3 +72,13 @@ metadata:
...
   secured.txt.url: "http://dummy-server/secured"
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: url-configmap-binary-download
+  labels:
+    findme: "yup"
+binaryData:
+  # Base64 encoded url is 'https://github.com/kiwigrid/k8s-sidecar/raw/master/test/kubelogo.png'
```
jekkel, line 83: I'd have preferred to target the dummy server (written in python) for the test case as this adds a dependency to an external service to every test run. But since we're running on github and if github is down the run is somewhat useless anyway this would be ok for me.

jekkel (dismissed):

jekkel (approved):
LGTM

jekkel:
```
Run cp test/kubelogo.png test/server/static/
cp: cannot stat 'test/kubelogo.png': No such file or directory
Error: Process completed with exit code 1.
```
see [here](https://github.com/kiwigrid/k8s-sidecar/actions/runs/3966064004/jobs/6797351884)

svolland-csgroup:
Not sure about this fix, i'm completely new to github workflows...

.github/workflows/build_and_test.yaml
```diff
@@ -6,6 +6,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
```
jekkel, line 9: I don't think we need this here - any specific reason for it?

jekkel: btw: it would be fine for me if we'd simply duplicate the png file in the server folder. Or use a symbolic link within the repository.
I think the docker build action builds everything via buildx and hence remotely and therefore clones the repo within the build action itself/

jekkel: We could also just move the png into the server folder and refer to it from the assertion scripts.

svolland-csgroup: I'll give it a last try, seems that only the static/ dir is missing (had it locally but forgot git does not push empty dirs), 

> I don't think we need this here - any specific reason for it?
buildx by default runs on a fresh git checkout, so you can't modify the build context with previous steps (here adding the png image).
Checking out and specifying the context path explicitly allows to circumvent this limitation.

PR #262, opened 2023-03-10
Allow skipping TLS verification for http requests

### LINKED ISSUE #261: Allow skipping TLS verification for http requests (not only to K8s API server)
Right now the SKIP_TLS_VERIFY env only has effect on the K8s API server connection.

However, the requests to Grafana cannot be customized. Even though there used to be a way to circumvent this in the past (which no longer works -- https://github.com/psf/requests/issues/6071), there should be an explicit way to skip TLS verification for Grafana (either reusing the same env SKIP_TLS_VERIFY or a separate one).
Since this runs as a sidecar, the hostname (localhost) won't match even with proper CA.

Ref:
https://github.com/kiwigrid/k8s-sidecar/blob/master/src/helpers.py#L130

lorenzo-biava:
Fixes #261


README.md
```diff
@@ -80,6 +80,8 @@ If the filename ends with `.url` suffix, the content will be processed as a URL
...
 | `ERROR_THROTTLE_SLEEP`     | How many seconds to wait before watching resources again when an error occurs                                                                                                                                                                                                                                                       | false    | `5`                                       | integer |
 | `SKIP_TLS_VERIFY`          | Set to `true` to skip tls verification for kube api calls                                                                                                                                                                                                                                                                           | false    | -                                         | boolean |
+| `SKIP_GRAFANA_TLS_VERIFY`          | Set to `true` to skip tls verification for Grafana api calls (for reloading certain resources).
```
jekkel, line 83: Please don't mention Grafana here. I understand the sidecar is especially popular for usage with grafana, but it is certainly not limited to it. This setting affects all http requests made based on `REQ_URL` config and for `*.url` triggered requests. So I suggest naming it `REQ_SKIP_TLS_VERIFY` then and document the affected calls (similar to `REQ_PASSWORD` I think).

jekkel (changes requested):
Thanks for your contribution! Looks good overall, only the name of the config needs to be changed.

Any chance you could add a test case as well?

lorenzo-biava:
@jekkel I added a Pod in the test resources, which uses that flag. It's not actually testing the HTTPS connection, but the logs should display the validation being skipped.

While we could add an actual HTTPS server to the testing scenario, I'm not sure it would be worth the extra effort in setting it up just for testing this flag...

jekkel:
@lorenzo-biava

Unfortunately the extra pod is almost useless unless there's at least a single assertion, don't you think?
So we either make this an actual regression test or we can avoid the confusion and omit the extra pod.
Thanks for renaming the config flag though!

jekkel (approved):

PR #263, opened 2023-03-16
Remove extra space from REQ_SKIP_TLS_VERIFY env

### LINKED ISSUE #261: Allow skipping TLS verification for http requests (not only to K8s API server)

lorenzo-biava:
Fixes #261 and #262, by removing the foreign space in the env variable


lorenzo-biava:
@jekkel sorry for the typo. Shouldn't have trusted copy-pasting here 😅

jekkel (approved):

jekkel:
your are welcome :)

Issue #268, opened 2023-04-03
UnboundLocalError: cannot access local variable 'dest_folder' where it is not associated with a value - 1.22


mkjmdski:
While upgrading image version on helm deployment of `grafana` (this specific value -> https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L774), with given config:

```yaml
        sidecar:
          image:
            repository: quay.io/kiwigrid/k8s-sidecar
            tag: 1.22.4

          resources:
            limits:
              cpu: 100m
              memory: 128Mi
            requests:
              cpu: 100m
              memory: 128Mi

          dashboards:
            # use when upgrading version of sidecar to spot issues
            # logLevel: DEBUG
            enabled: true
            label: grafana_dashboard
            folderAnnotation: grafana_dashboard_dir
            provider:
              foldersFromFilesStructure: true

          datasources:
            # use when upgrading version of sidecar to spot issues
            # logLevel: DEBUG
            enabled: true
            label: grafana_datasource
            skipReload: true
            initDatasources: true

```

last working value was: `1.19.2`. Probably it's regression from 1.22 -> https://github.com/kiwigrid/k8s-sidecar/pull/231

anonymized stacktrace from DEBUG log of init-container grafana-init-sc-datasources. All of our datasources are secretes configured via ExternalSecret in `extraObjects:` key. They all contain `k8s-sidecar-target-directory: "."` annotation.

```log
{"time": "2023-04-03T14:24:50.726994+00:00", "msg": "Working on secret: default/grafana-datasource", "level": "DEBUG"}
{"time": "2023-04-03T14:24:50.727101+00:00", "msg": "Found a folder override annotation, placing the grafana-datasource in: /etc/grafana/provisioning/datasources/.", "level": "INFO"}
{"time": "2023-04-03T14:24:50.727705+00:00", "msg": "Writing azure.yaml", "level": "DEBUG"}
{"time": "2023-04-03T14:24:50.727894+00:00", "msg": "Contents of azure.yaml haven't changed. Not overwriting existing file", "level": "DEBUG"}
... x 7 for all of the secrets
{"time": "2023-04-03T14:24:50.820574+00:00", "msg": "response body: {\"kind\":\"ConfigMapList\",\"apiVersion\":\"v1\",\"metadata\":{\"resourceVersion\":\"586142921\"},\"items\":[]}\n", "level": "DEBUG"}
Traceback (most recent call last):
[...]

PR #281, opened 2023-05-17
Fix issue 265 and 268 and improve logging

### LINKED ISSUE #265: when listing secrets and configmaps, secrets fail with 'V1Secret' object has no attribute 'binary_data'
We have a mix of configmaps and secrets for grafana datasources. When the sidecar tries to load the secrets it fails. Neither the configmaps nor the secrets use binary data:

`Traceback (most recent call last):
  File "/app/sidecar.py", line 155, in <module>
    main()
  File "/app/sidecar.py", line 113, in main
    list_resources(label, label_value, target_folder, request_url, request_method, request_payload,
  File "/app/resources.py", line 128, in list_resources
    files_changed |= _process_config_map(dest_folder, item, resource, unique_filenames, enable_5xx, True)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/resources.py", line 217, in _process_config_map
    if config_map.binary_data is not None:
       ^^^^^^^^^^^^^^^^^^^^^^
 AttributeError: 'V1Secret' object has no attribute 'binary_data'`
       
We tried running with the latest version 1.23.1 but that didnt help.

mario-steinhoff-gcx:
This PR fixes #265 and #268.

I also improved logging a bit because I had a hard time understanding the control flow.

- an INFO log line is printed when a list sync is started
- a DEBUG log line is printed when a watch sync is started (INFO seems way too spammy with a default timeout of 60s)
- an INFO log line is printed when a file is written or removed to disk - previously only a debug log line was printed without the full absolute path and even if the actual file was not changed
- an ERROR log line is printed instead of `printf()` when a file cannot be removed

I built changes in a local image and it seems to work again.


tomrk-esteam8:
Hi, thank you for contribution. That's great. Could you please add as well some tests including for example secret and config map with the same name?

mario-steinhoff-gcx:
@tomrk-esteam8 I added some tests as requested, but I'm not sure how to run them locally and they need to be approved first to run in GHA.

Currently there are no tests for the `LIST` method, but I'm not sure how to design a test for that.

tomrk-esteam8 (approved):
Seems to be fine for me, when tests will work, we can merge @jekkel

mario-steinhoff-gcx:
No idea why the test fails.

To to pinpoint the issue, I modified the workflow to also upload artifacts with all the expected files after first sync. I also split up the tests by `sidecar`/`sidecar-5xx`/`sidecar-pythonscript`.

mario-steinhoff-gcx:
I forgot an `mkdir` and pushed a fix, need approval again.

mario-steinhoff-gcx:
It looks like the problem was in the pythonscript test. The test compares the number of `Hello from python script!` occurences and currently it expects 7 occurences but in the logs we have 9, probably because I added 2 more files. I fixed the test, lets see if it works now.

tomrk-esteam8:
ok, I am just approving tests today for you, will chec it more detail when it works tomorrow, thank you for your work here :)

mario-steinhoff-gcx:
> ok, I am just approving tests today for you

May I suggest to redesign the tests so they can be run on the local machine without a dependency on the GHA workflow? It makes sense to require approval for changes to the GHA workflow on approval but, right now it seems to be the only way to run the the tests and because the tests are embedde in the GHA workflow config, every little change needs a new approval.

Tests failed again but the good news is that the pythonscript test works now, and it seems to be just a copy-paste error I made. Pushed another fix.

jekkel (approved):

jekkel:
> May I suggest to redesign the tests so they can be run on the local machine without a dependency on the GHA workflow? It makes sense to require approval for changes to the GHA workflow but right now it seems to be the only way to run the the tests. And because the tests are embedded in the GHA workflow config every little change needs a new approval.

That's something we wished for quite some time already. I was thinking about using some kind of regular python test framework such that there's some tool support at least. Do you have any specific suggestions? Can you imagine to contribute this?

PR #285, opened 2023-05-24
try to parse `REQ_PAYLOAD` as json to avoid unexpected quoting

fbrousse:
the post method converts the json to string. Here we read from a os.environ and thus a string. So we send a string converted to string. It adds extra quotes and breaks for some server implementation (seen with grafana for provisioning). With the fix here we convert first the payload to dict to get it in the right format.

### DIFF
--- a/src/helpers.py
+++ b/src/helpers.py
@@ -109,7 +109,6 @@ def remove_file(folder, filename):
         logger.error(f"Unable to remove {complete_file}, file not found")
         return False
 
-
 def request(url, method, enable_5xx=False, payload=None):
     enforce_status_codes = list() if enable_5xx else [500, 502, 503, 504]
 
@@ -139,7 +138,7 @@ def request(url, method, enable_5xx=False, payload=None):
     if method == "GET" or not method:
         res = r.get("%s" % url, auth=auth, timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
     elif method == "POST":
-        res = r.post("%s" % url, auth=auth, json=json.loads(payload), timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
+        res = r.post("%s" % url, auth=auth, json=payload, timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
     else:
         logger.warning(f"Invalid REQ_METHOD: '{method}', please use 'GET' or 'POST'. Doing nothing.")
         return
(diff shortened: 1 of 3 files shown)


fbrousse:
Extract from : https://reqbin.com/code/python/m2g4va4a/python-requests-post-json-example#:~:text=To%20post%20a%20JSON%20to,it%20to%20a%20JSON%20string.

**To post a JSON to the server using Python Requests Library, call the requests.post() method and pass the target URL as the first parameter and the JSON data with the json= parameter. The json= parameter takes a dictionary and automatically converts it to a JSON string.**

src/helpers.py
```diff
@@ -139,7 +139,7 @@ def request(url, method, enable_5xx=False, payload=None):
     if method == "GET" or not method:
         res = r.get("%s" % url, auth=auth, timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
     elif method == "POST":
-        res = r.post("%s" % url, auth=auth, json=payload, timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
+        res = r.post("%s" % url, auth=auth, json=json.loads(payload), timeout=REQ_TIMEOUT, verify=REQ_TLS_VERIFY)
```
jekkel, line 142: I think the fix is valid. However, the json parsing should be done initially when loading from the environment, not each time a request is being made.
Also - to be backwards compatible - when parsing fails we need to fall back to the current behavior.

jekkel (changes requested):
Thanks for your contribution! It would be great if you could follow up on the comments and maybe even provide a test case?

src/helpers.py
fbrousse: fair enough, I updated with a method doing the conversion and a fallback.

src/helpers.py
```diff
@@ -109,7 +109,6 @@ def remove_file(folder, filename):
         logger.error(f"Unable to remove {complete_file}, file not found")
         return False
 
-
```
jekkel, line 112: please re-add this line

src/resources.py
```diff
@@ -54,6 +55,14 @@ def signal_handler(signum, frame):
 
 signal.signal(signal.SIGTERM, signal_handler)
 
+def prepare_payload(payload):
+    """Prepare payload as dict for request."""
+    logger.debug(f"fbrousse testing.")
```
jekkel, line 60: I guess this shouldn't make it eventually...?

src/resources.py
```diff
@@ -54,6 +55,14 @@ def signal_handler(signum, frame):
...
+    try:
+       payload_dict = json.loads(payload)
```
jekkel, line 62: we could return early here, couldn't we?

jekkel (approved):

PR #294, opened 2023-07-05
fix test for execute permission of script file

macropin:
This PR resolves the following issues:

- Executable check does not correctly identify when the user has execute permission on the script (S_IXOTH only checks "execute/search permission, others" ie group, not the owner)
- Change log level of execute handler logger to info: resolves lack of visibility on script execution


pragmatic (approved):

jekkel (approved):
LGTM

PR #332, opened 2024-03-04
Stay on python stable

jkroepke:
It seems like python is NOT using semver for they images and dependabot is [too stupid](https://github.com/kiwigrid/k8s-sidecar/pull/323) and offers an Alpha version of python here.

Instead, dependabot will not update the alpine version. It seems like dependabot does not help here.

An potential alternative would be `python:3.12-alpine`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12.2-alpine3.19 as base
+FROM python:3.12-alpine as base
 
 FROM base as builder
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13.0a4-alpine3.18 as base
+FROM python:3.12.2-alpine3.19 as base
 
 FROM base as builder
 WORKDIR /app


ChristianGeie:
@jkroepke thx for the issue. For other Python projects in our company, we use the same base image, but we use Renovate for dependency management. At least there it doesn't happen that alpha versions are suggested.

But whatever, i think we can give `python:3.12-alpine` a try and then see how dependebot behaves.

Can you please create a PR for this right away?

ChristianGeie:
@jkroepke my fault, PR is already in place.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.13.0a4-alpine3.18 as base
+FROM python:3.12.2-alpine3.19 as base
```
jkroepke, line 1:
```suggestion
FROM python:3.12.2-alpine as base
```

jkroepke:
I hope its fine now.

ChristianGeie (approved):
lgtm

PR #333, opened 2024-03-07
Add exception_handler

jkroepke:


ChristianGeie:
@jkroepke so that I understand this correctly: You want to use the `sys.excepthook` function to set a global exception handler for uncaught exceptions, so this funtion is called via the custom exception hook `sys.excepthook = exception_handler` whenever an uncaught exception occurs.

jkroepke:
Yes

ChristianGeie:
thanks for confirming that. Meantime i extend the tests. pls rebase so i can trigger the workflow again.

jkroepke:
Done

ChristianGeie (approved):
lgtm

PR #356, opened 2024-08-16
#minor Add an option to send request only once per batch

ahtohio:
Reloading on each received event (with watch method) can be quite costly. I run Grafana provisioning and asking it to reload all my provisioned configmaps (>300) on each event is not needed and only increases cpu load.

Reload once after each WATCH_SERVER_TIMEOUT seconds should be enough.

I've added an option REQ_ONCE_PER_BATCH that can toggle this reload-once-per-watch-cycle reload behaviour, false by default to not interfere with existing logic.


ahtohio:
Hi, @ChristianGeie 
I've added some tests related to this functionality. For some reason, at the last commit all jobs failed for quite unknown reasons, but new workflow on the same commit in my personal branch worked well.
But I can't restart/create new workflow in this repo, it seems

ChristianGeie:
thx @ahtohio . I do a lot of changes regarding workflow actions in the last couple of weeks, please excuse that sometimes the tests didn't work reliably.
Otherwise your changes look pretty good to me.

isubasinghe:
@ChristianGeie this is a pretty important fix in my opinion, my 12 core CPU gets hammered while running our application.


If there is anything I can do to progress this PR, please let me know.

jcpunk:
Can you rebase this off HEAD?

ChristianGeie (changes requested):
pls rebase to `master`

PR #365, opened 2024-09-27
update kind node images

ChristianGeie:
add comment to point out used kind version out of kind-action

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -1,3 +1,5 @@
+---
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
 name: "Build and Test"
 on:
   - pull_request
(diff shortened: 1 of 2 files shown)


ChristianGeie:
@jekkel @tomrk-esteam8 need review pls.

.github/workflows/build_and_test.yaml
```diff
@@ -43,27 +45,27 @@ jobs:
           - maj_min: v1.22
             digest: sha256:f5b2e5698c6c9d6d0adc419c0deae21a425c07d81bbf3b6a6834042f25d4fba2
```
jekkel, line 46: kind 0.22.0 doesn't seem to officially support anything below 1.23 anymore. Shouldn't we delete these then?

jekkel (approved):

.github/workflows/build_and_test.yaml
ChristianGeie: Thx for pointing out. I agree to remove v1.21 and v1.22

ChristianGeie:
I will also adjust tests as required checks in the repo config

tomrk-esteam8 (approved):
LGTM

PR #399, opened 2025-06-23
Logging improvements

andrew-pickin-epi:
* Sending URL request need a log Info event: Requiring debug log level to see url request is either unhelpful or too noisy. Also mirrors script running log event.
* When logging of url request "None" was reported if defaulting to GET method.
* Useful to see payload in event on POST.
* Removal of ellipsis in config load.
* A missing, non-mandatory, option (folder annotation) shouldn't result in a warning.


ChristianGeie (approved):
lgtm, thx for contribution

PR #425, opened 2025-10-07
Fix file deletes in SLEEP mode

### LINKED ISSUE #499: error/issue on initial sync - works perfectly fine for changes after the initial sync
Hi kiwigrid! 

I'm seeing an error on startup of the sidecar (I'm using it together with grafana). At initial sync the request is not properly sent in this case. If there are changes to the monitored components later on though it works perfectly fine..

My env-vars are these ones:

```
- name: HEALTH_PORT
  value: '8080'
- name: REQ_SKIP_TLS_VERIFY
  value: 'true'
- name: METHOD
  value: WATCH
- name: LABEL
  value: grafana_datasource
- name: FOLDER
  value: /etc/grafana/provisioning/datasources
- name: RESOURCE
  value: secret
- name: RESOURCE_NAME
  value: secret/grafana-datasources
- name: NAMESPACE
  value: 'namespace,different-namespace'
- name: REQ_USERNAME
  value: 'IWouldNever'
- name: REQ_PASSWORD
  value: 'WriteThisHere'
- name: REQ_URL
  value: 'https://localhost:3000/api/admin/provisioning/datasources/reload'
- name: REQ_METHOD
  value: POST
- name: REQ_SKIP_INIT
  value: 'true'

```

The initial write works but then I see an error that apparently stops the request:

```
{"time": "2025-12-26T13:22:24.491495+00:00", "level": "INFO", "msg": "Writing /etc/grafana/provisioning/alerting/some_alert_name.yaml (ascii)"}
[...]
{"time": "2025-12-26T13:22:24.492460+00:00", "level": "ERROR", "msg": "Error when updating from 'some_alert_name.yaml' into 'None'", "exc_info": "Traceback (most recent call last):\n File \"/app/resources.py\", line 332, in _update_file\n return remove_file(dest_folder, filename)\n File \"/app/helpers.py\", line 110, in remove_file\n complete_file = os.path.join(folder, filename)\n File \"<frozen posixpath>\", line 77, in join\nTypeError: expected str, bytes or os.PathLike object, not NoneType"}
```

The error happens as the `_update_file` method is called with `None` as the `dest_folder` variable content. I haven't quite figured out though why it is implemented like this. As far as I understood the `is_removed` variable in both `_process_config_map` and `_process_secret` actually defines whether a file gets removed or not. 

I found the initall call of the method with `None` as `dest_folder` here:

https://github.com/kiwigrid/k8s-sidecar/blob/b619601a21507b87acffee8ca45e1a91bf21b53d/src/resources.py#L176-L185

and I created a "Fix" which sets the the proper and existing `dest_folder` (using `_get_destination_folder` method) just like in the for item in items loop above. With this fix the error doesn't happen like that.

Unfortunately I couldn't reproduce the error in any different k8s setting but need to work with the existing one. However I tried setting the `dest_folder` even for those 

If I missed something please let me know and/or close the PR for it - that PR however seems to work for our case..

madaraszg-tulip:
When running k8s-sidecar in sleep mode if a configmap or secret is removed, k8s-sidecar logs the following message:

```
{"time": "2025-10-07T10:58:13.578900+00:00", "level": "ERROR", 
"msg": "Error when updating from 'mimir-custom-alerts.yaml' into 'None'", 
"exc_info": "Traceback (most recent call last):\n  File \"/app/resources.py\", line 329, in _update_file\n
    return remove_file(dest_folder, filename)\n  File \"/app/helpers.py\", line 110, in remove_file\n
    complete_file = os.path.join(folder, filename)\n  File \"<frozen posixpath>\", line 77, in join\n
TypeError: expected str, bytes or os.PathLike object, not NoneType"}
```

This #patch should fix the issue by using `old_dest_folder` instead of `dest_folder` when we're removing files.

### DIFF
--- a/src/resources.py
+++ b/src/resources.py
@@ -194,6 +194,7 @@ def _process_secret(dest_folder, secret, resource, unique_filenames, enable_5xx,
     old_secret = _resources_object_map[resource].get(secret.metadata.namespace + secret.metadata.name) or copy.deepcopy(secret)
     old_dest_folder = _resources_dest_folder_map[resource].get(secret.metadata.namespace + secret.metadata.name) or dest_folder
     if is_removed:
+        dest_folder = old_dest_folder
         _resources_object_map[resource].pop(secret.metadata.namespace + secret.metadata.name, None)
     else:
         _resources_object_map[resource][secret.metadata.namespace + secret.metadata.name] = copy.deepcopy(secret)
@@ -234,6 +235,7 @@ def _process_config_map(dest_folder, config_map, resource, unique_filenames, ena
     old_config_map = _resources_object_map[resource].get(config_map.metadata.namespace + config_map.metadata.name) or copy.deepcopy(config_map)
     old_dest_folder = _resources_dest_folder_map[resource].get(config_map.metadata.namespace + config_map.metadata.name) or dest_folder
     if is_removed:
+        dest_folder = old_dest_folder
         _resources_object_map[resource].pop(config_map.metadata.namespace + config_map.metadata.name, None)
     else:
         _resources_object_map[resource][config_map.metadata.namespace + config_map.metadata.name] = copy.deepcopy(config_map)


ChristianGeie:
Hi @madaraszg-tulip 
first thx for contribution.
By adding `dest_folder = old_dest_folder` when an item is removed we better ensure that the correct folder is used for file deletion, even if `None` was passed to `_process_secret` or `_process_config_map`.

madaraszg-tulip:
> Hi @madaraszg-tulip first thx for contribution. By adding `dest_folder = old_dest_folder` when an item is removed we better ensure that the correct folder is used for file deletion, even if `None` was passed to `_process_secret` or `_process_config_map`.

Do I need to do anything else with this PR to get it reviewed?

CarstenSon:
nvm, i didn't check properly :D 

@ChristianGeie : is there anything elsle to do for this PR? 

I'm running into the same issue and it looks like this is the fix towards it..

madaraszg-tulip:
@ChristianGeie do you need anything else from me for this PR?

ChristianGeie (approved):
Apologies for the delay in reviewing, I unfortunately lost track of this PR.

The fix is precisely correct. Using `old_dest_folder` within the `is_removed` block cleanly resolves the `TypeError` and ensures that files are deleted from the correct directory.

Thx, lgtm

ChristianGeie (approved):
lgtm

ChristianGeie:
@dependabot rebase

ChristianGeie:
It looks like the required CodeQL check is stuck. i will close and reopen this PR to re-trigger the CI checks

PR #433, opened 2025-10-16
wyn_skip_init: - add flag to skip initial request to REQ_URL when using WATCH

### LINKED ISSUE #432: Native sidecar compatibility
https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

What I'd like to do is use k8s-sidecar to load all of the configmaps/secrets initially, _then_  start the main container. I'd also like it to WATCH and reload the main container.

The way to go about this is to use native sidecars with a startupProbe

However currently, if a REQ_URL is specified to reload the main container, then the sidecar will never become ready, as the main container is waiting for the sidecar to initialize

k8s-sidecar should provide some mechanism to avoid calling REQ_URL on the initial sync while in WATCH mode

sthomson-wyn:
Fixes #432


ChristianGeie:
@sthomson-wyn Thanks a lot for your contribution and for addressing this startupProbe / REQ_URL initialization issue. 
Your approach makes total sense and aligns well with how native sidecars behave in Kubernetes.

Could you please rebase your branch on the latest `master`?
There have been several updates and refactorings recently (including watcher and initialization changes), and we’d like to make sure your patch applies cleanly before review and merge.

sthomson-wyn:
@ChristianGeie updated, thanks

Issue #468, opened 2025-11-19
Please use immutable tags


sathieu:
We use renovate to pin images with digests. 

The tag [2.1.2](https://quay.io/repository/kiwigrid/k8s-sidecar?tab=tags&tag=2.1.2) changed its digest from `sha256:8b875915be0869624cf5dfd48a1dfbfb627e778793cc1612d00fb281a8805c63` (few days ago) to `sha256:5304aa2b45f80a19f84a5e9dfb139c9caeaeecf1f9e507d0bdb9bc1cdb87a58a`.

Please unsure the tag doesn't change, or bump the tag version if you need to rebuild images.

ChristianGeie:
Thank you very much. Of course, the digest for a version should not change, and it is always my intention not to do so. I have no idea why this happened.

ChristianGeie:
My changes does not work:

k8s-sidecar:2.1.3 was build using disgest 53c71a4b67c692074ca7de42e0296c45ff6305348645e451678fb92f334966d5
After merging of #471 and #472 into master without any tag in merge commit a release v2.1.3 using `digest:305f3a955a448b27ee4be7777e1c3fceb69b1ba2af3db24a392a02e79e1658b3` was build.

ChristianGeie:
I will close this issue. With all adjustments out of #475 changes on the codebase should trigger a new build every time an increase the artifact version.

PR #473, opened 2025-11-25
Bump using minor version on all merges

sathieu:
Fixes: #468

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,7 +23,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DEFAULT_BRANCH: master
           INITIAL_VERSION: 1.0.0
-          DEFAULT_BUMP: none
           BRANCH_HISTORY: last
       - name: "Set up QEMU"
         if: steps.tagging.outputs.part


ChristianGeie:
Thx for contributing. 
Removal of `DEFAULT_BUMP: none` that doesn't solve the actual problem - it just shifts it.

I think the following typically happens:
* The action default is `patch`
* Every push to `master` without a special marker (#major, #minor, #patch, #none, etc.) then automatically creates a new tag (e.g., from 2.12.2 to 2.12.3)
* Every subsequent steps (build, push, release) will then always run because a new tag is always created
* This would solve the problem of “old tag gets new digest” (because there is always a new tag),
but: i also get releases for pure CI/docs/meta changes -> probably not what you or i want

The actual goal is:
* Only release if you deliberately set a bump keyword (#patch, #minor, #major)
* Do not release for pure CI/dependency/meta changes

`DEFAULT_BUMP: none` is just right for this - but i think i need to adjust the `IF` condition:

`if: steps.tagging.outputs.part`

instead i need to adjust this to

`if: steps.tagging.outputs.new_tag != ''`

Therefore:
* No keyword + `DEFAULT_BUMP: none` -> no `new_tag` -> no build, no release
* Keyword in commit -> `new_tag` set -> build, push, release run

So: Leaving `DEFAULT_BUMP: none` in place, but link all relevant steps to `new_tag`. What do you think?

ChristianGeie:
I’ve taken over your changes, rebased them on the latest master and opened a new PR: #475 .  
We’ll continue the review and merge process over there.

Closing this one in favour of #475 .

Issue #427, opened 2025-10-13
No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/app']


cjedro:
When upgrading from 1.30.10 -> 1.30.11 in Grafana Helm chart i get the following error:

```json
{
   "time":"2025-10-13T09:23:21.882413+00:00",
   "level":"ERROR",
   "msg":"FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/app']",
[...]

PR #354, opened 2024-07-12
Bumps python from 3.12-alpine to 3.20-alpine

ChristianGeie:

PR #489, opened 2025-12-11
fix: pin python version to 3.13.11 in dockerfile

mjnagel:
Pins the image to a specific python version (3.13.11) and bumps to the newer alpine 3.23. Also pins to a digest.

This got "unpinned" in [this PR](https://github.com/kiwigrid/k8s-sidecar/pull/354), potentially by accident since the python and alpine tags were similar?

Related to https://github.com/kiwigrid/k8s-sidecar/issues/427

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.14.2-alpine3.23 AS base
+FROM python:3.13.11-alpine3.23@sha256:51b5354ed44df6e1a3b3faf6d3a3d40da129621046b6d5a707b7c1f44d258ed6 AS base
 
 FROM base AS builder
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:alpine3.22 AS base
+FROM python:3.14.2-alpine3.23 AS base
 
 FROM base AS builder
 WORKDIR /app


ChristianGeie:
@mjnagel thx for contributing

Pinning Python and Alpine makes sense.
One question though: should we consider pinning the image by digest as well (e.g. `python:3.14.2-alpine3.23@sha256:...`) to make builds fully reproducible and avoid silent upstream rebuilds?

Since Dependabot is already managing dependencies in this repo, it could also handle digest updates automatically.
Curious what you think.

ChristianGeie (changes requested):
python 3.14 is really very new, so my fear is that dependencies like kubernetes, requests or urllib3 breaks the build. I suggest to use [3.13.11-alpine3.23](https://hub.docker.com/layers/library/python/3.13.11-alpine3.23/images/sha256-5e93b66c7d2672badb5c780b2030e6dc641ae93130b2c9f66851f47b462fd74e) for max. stability

mjnagel:
> One question though: should we consider pinning the image by digest as well (e.g. `python:3.14.2-alpine3.23@sha256:...`) to make builds fully reproducible and avoid silent upstream rebuilds?

Definitely makes sense to pin to a digest as well - I'll do that.

> python 3.14 is really very new, so my fear is that dependencies like kubernetes, requests or urllib3 breaks the build. I suggest to use [3.13.11-alpine3.23](https://hub.docker.com/layers/library/python/3.13.11-alpine3.23/images/sha256-5e93b66c7d2672badb5c780b2030e6dc641ae93130b2c9f66851f47b462fd74e) for max. stability

I'm okay with this since it is a newer release. I think from what I've seen there were a few unexpected issues with the new builds, related to 3.14 (the linked one and the memory usage increase). I'll go ahead and make those changes.

mjnagel:
@ChristianGeie any chance this could get re-reviewed (and maybe https://github.com/kiwigrid/k8s-sidecar/pull/490 get released as well)? There's the related open issues for memory consumption and writable file requirements, but also some downstream issues with consumers of it (Grafana for example: https://github.com/grafana/helm-charts/issues/4031). Thanks!

ChristianGeie (dismissed):
lgtm

ChristianGeie:
feature added in #552

PR #500, opened 2025-12-26
fix: use dest_folder instead of None to remove cache that is not listed

CarstenSon:
this would fix the issue presented in #499


CarstenSon:
just found https://github.com/kiwigrid/k8s-sidecar/pull/425 and saw that that solution is probably better than this one..

PR #505, opened 2026-01-06
fix(healthz): Correct uvicorn log propagation for health server

ChristianGeie:
feat(logging): Unify logging to structured JSON format

Refactors the application's logging to ensure all output, including
from the Uvicorn web server, is in a structured JSON format. This aligns
with modern Kubernetes practices and fulfills the goal of having uniform
log messages.

- Removes the separate plain-text 'uvicorn_simple' formatter from
  `healthz.py`.
- Configures all `uvicorn` loggers (`uvicorn`, `uvicorn.error`,
  `uvicorn.access`) to use the main `console` handler, which is
  configured with `JsonFormatter`.
- Disables propagation for uvicorn loggers to prevent duplicate
  messages and conflicts with the root logger.
- Adds a logging filter to remove the unnecessary `color_message` field
  from Uvicorn's log output, keeping the JSON logs clean.

This change resolves inconsistencies in the log output and fixes the
underlying issue that caused the `Verify health server startup` test to
fail, while ensuring the sidecar's LOG_LEVEL is respected by Uvicorn.

### DIFF
--- a/src/logger.py
+++ b/src/logger.py
@@ -55,6 +55,17 @@ class JsonFormatter(Iso8601Formatter, jsonlogger.JsonFormatter):
         super(JsonFormatter, self).add_fields(log_record, record, message_dict)
 
 
+class RemoveColorMessageFilter(logging.Filter):
+    """
+    A logging filter that removes the 'color_message' attribute from log records.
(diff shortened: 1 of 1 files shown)


ChristianGeie (commented):
checks are green

PR #511, opened 2026-01-16
Ensure health server supports dual-stack

ChristianGeie:
fix(healthz): Ensure health server supports dual-stack
- modifies the health check server to correctly listen on both IPv4 and IPv6 interfaces. This is achieved by explicitly setting the address family to AF_INET6 on the HTTPServer, which enables dual-stack support on compliant systems.
- Additionally, the test workflow has been updated to include a specific test that verifies the health endpoint's accessibility over IPv4 and IPv6, ensuring the fix is properly validated.

feat(tests): Enable dual-stack networking in kind
- This enables IPv4/IPv6 dual-stack networking for the Kubernetes-in-Docker (kind) cluster used in the test workflow.

docs: Mention dual-stack support for health endpoint
- updates README to clarify that the /healthz endpoint is compatible with both IPv4 and IPv6 environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ If the filename ends with `.url` suffix, the content will be processed as a URL 
 
 ## Health Endpoint
 
-The sidecar provides a health endpoint at `/healthz` on port `8080` (or as configured by `HEALTH_PORT`) that can be used for Kubernetes readiness and liveness probes.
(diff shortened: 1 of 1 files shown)


ChristianGeie:
Hi @AshaRajput-27 and @jimmy-ungerman,

Thank you both for your work on this issue.

Building on the initial fix in PR #510, I've pushed some additional commits to the fix/510-dual-stack-test branch. These changes complete the dual-stack implementation for the health endpoint and add the necessary tests to validate it in a dual-stack cluster environment.

We would appreciate it if you could review the new changes in this pull request.

Thanks

AshaRajput-27:
Hi @ChristianGeie,
Thank you for sharing this. I’ve reviewed the changes and they look good to me.

PR #538, opened 2026-03-27
feat(auth): add support for HTTPBearerAuth using Kubernetes service account

### LINKED ISSUE #321: Support jwt token auth
As more secure alternative to auth basic, add the support via jwt token.

it should possible to define a token file and a header name to attach the token. The token file can be changed each hour. 

Use case:

i have to disable the initial admin in grafana. I would like to enable jwt auth to allow login via kubernetes server account token, since its a short-living jwt token.

ref: https://github.com/grafana/grafana/issues/81794

jkroepke:
Fixes #321


ChristianGeie:
Hi @jkroepke thx for contribution. This is an absolutely useful and contemporary feature. Regardless of the changes I have already suggested above, I would like to have a test case for the feature:
Maybe a definition of a new Test-Pod in test/resources/sidecar.yaml and a new check in build_and_test.yaml something like this: 
```
      - name: Verify Service Account Authentication
        shell: bash
        run: |
          source /tmp/sidecar_helpers.sh
          echo "--- Verifying Service Account Authentication ---"
          # We just need to check if the dummy-server received the authenticated request.
          wait_for_pod_log "dummy-server-pod" "TEST_SUCCESS: Received Authorization header with Bearer token."
```

Thx

README.md
```diff
@@ -83,6 +83,8 @@ If the filename ends with `.url` suffix, the content will be processed as a URL
...
+| `REQ_AUTH_METHOD`          | Authentication method to use for requests to `REQ_URL` and for `*.url` triggered requests. Options are `basic` for basic authentication, `jwt_token_file` for authentication with jwt token from a file and `none` for no authentication. If not set, it will be determined based on the presence of `REQ_USERNAME`/`REQ_PASSWORD` or their file-based counterparts. | false | - | string  |
+| `REQ_AUTH_TOKEN_FILE` | Path to file containing the token to use for authentication with `jwt_token_file` method for requests to `REQ_URL` and for `*.url` triggered requests. This overrides the `REQ_AUTH_METHOD` if set, meaning that if this variable is set, the sidecar will use the token in this file for authentication regardless of the value of `REQ_AUTH_METHOD`. | false    | `/var/run/secrets/kubernetes.io/serviceaccount/token` | string  |
```
ChristianGeie: You states that `REQ_AUTH_TOKEN_FILE` overrides `REQ_AUTH_METHOD`. However, the implementation in src/helpers.py first checks `REQ_AUTH_METHOD` and ignores `REQ_AUTH_TOKEN_FILE` if the method is not jwt_token_file.
```suggestion
| REQ_AUTH_TOKEN_FILE | Path to file containing the token. This is only used when REQ_AUTH_METHOD is set to jwt_token_file. |
  false    | /var/run/secrets/kubernetes.io/serviceaccount/token | string  |
```

ChristianGeie:
Currently failing tests points me to a new issue: The logs show that authentication itself is working. The sidecar sends a request to `http://dummy-server/secured-bearer` and receives a successful `200 OK allowed` response.
The problem: The k8s watch expires after 60 seconds, which is normal behavior. When the watch is restarted, the `url-configmap-jwt` ConfigMap is reloaded as an ADDED event. As a result, the request is sent to the dummy-server again, even though nothing has changed.
The sidecar should not fetch the URLs again every time the watch process is refreshed, but only when the content of the resource has actually changed.

I think the problem lies in the order of operations. When a resource is processed (e.g., after a watch reconnect), the code performs the following steps:

1. The `_iterate_data` function iterates through all entries in the ConfigMap
2. For the entry `secured-bearer.txt.url`, `_update_file` is called
3. Within `_update_file`, `_get_file_data_and_name` is called
4. Here, the HTTP request to the URL is executed immediately to retrieve the content
5. Only then, in `write_data_to_file`, is the new content compared with the old content on the hard drive, and it is determined that nothing has changed

The request is therefore sent before checking whether the URL in the ConfigMap has changed at all.

jkroepke:
tbh. no idea what I had todo here. It looks like you have a plan and `Allow edits and access to secrets by maintainers` is enabled. Would you able to resolve this?

PR #573, opened 2026-05-10
fix: scope the diff for multi-namespaced setups

### LINKED ISSUE #572: Multi-namespace SLEEP-mode sync deletes resources from other configured namespaces
### Description

When `NAMESPACE` is set to a comma-separated list (e.g. `NAMESPACE=example-ns,second-ns`) and the sidecar runs in `METHOD=LIST` / SLEEP mode (or any path that forces SLEEP, such as `RESOURCE_NAME` being set), files synced from one namespace get repeatedly deleted by the watcher thread of another namespace.

### Reproduction

1. Set `NAMESPACE=example-ns,second-ns`, `METHOD=LIST` (or `RESOURCE_NAME=...`).
2. Place a labeled ConfigMap in each namespace.
3. Observe that on each SLEEP cycle, the file from one namespace is removed and re-created in alternation.

CarstenSon:
fixes #572

### DIFF
--- a/src/resources.py
+++ b/src/resources.py
@@ -189,11 +189,18 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
         if resource == RESOURCE_CONFIGMAP:
             files_changed |= _process_config_map(dest_folder, item, resource, unique_filenames, enable_5xx)
         else:
-            files_changed = _process_secret(dest_folder, item, resource, unique_filenames, enable_5xx)
+            files_changed |= _process_secret(dest_folder, item, resource, unique_filenames, enable_5xx)
 
-    # Clear the cache that is not listed.
-    for key in set(_resources_object_map[resource].keys()) - exist_keys:
-        item = _resources_object_map[resource].get(key)
+    # Clear the cache that is not listed. Scope the diff to this namespace: the cache is shared across per-namespace threads, so an unscoped diff would let one thread delete another's resources. 
+    resource_objects = _resources_object_map[resource].copy()
+    relevant_keys = {
+        key for key, item in resource_objects.items()
+        if namespace == "ALL" or item.metadata.namespace == namespace
+    }
+    for key in relevant_keys - exist_keys:
+        item = resource_objects.get(key)
+        if item is None:
+            continue  # another thread has already removed the key
         metadata = item.metadata
 
         logger.debug(f"Removing {resource}: {metadata.namespace}/{metadata.name}")
@@ -201,7 +208,7 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
         if resource == RESOURCE_CONFIGMAP:
             files_changed |= _process_config_map(None, item, resource, unique_filenames, enable_5xx, True)
         else:
-            files_changed = _process_secret(None, item, resource, unique_filenames, enable_5xx, True)
+            files_changed |= _process_secret(None, item, resource, unique_filenames, enable_5xx, True)
 
     if script and files_changed:
         execute(script)
--- a/src/resources.py
+++ b/src/resources.py
@@ -191,8 +191,12 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
         else:
             files_changed = _process_secret(dest_folder, item, resource, unique_filenames, enable_5xx)
 
-    # Clear the cache that is not listed.
-    for key in set(_resources_object_map[resource].keys()) - exist_keys:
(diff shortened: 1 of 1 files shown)


src/resources.py
```diff
@@ -191,8 +191,12 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
         else:
             files_changed = _process_secret(dest_folder, item, resource, unique_filenames, enable_5xx)
 
-    # Clear the cache that is not listed.
-    for key in set(_resources_object_map[resource].keys()) - exist_keys:
+    # Clear the cache that is not listed. Scope the diff to this namespace: the cache is shared across per-namespace threads, so an unscoped diff would let one thread delete another's resources. 
+    relevant_keys = {
```
ChristianGeie, line 196: We should read/change the cache using a lock or at least copy it in a safe way:

```
resource_objects = _resources_object_map[resource].copy()
relevant_keys = {
    key for key, item in resource_objects.items()
    if namespace == "ALL" or item.metadata.namespace == namespace
}
```

ChristianGeie (changes requested):
The namespace filtering itself is technically correct and resolves #572, but it still needs protection against concurrent cache changes.

CarstenSon:
@ChristianGeie I updated the PR accordingly - forgot about using a copy before..

thank you!

ChristianGeie:
After thinking a lot about this, here are my thoughts:

1. The `.copy()` method only protects the construction of `relevant_keys`. The loop body continues to read from the live dictionary that has not been copied. Between the snapshot and the `.get(key)` call, another thread (a WATCH thread with a DELETED event or a second LIST thread) can pop the same key via `_process_secret(is_removed=True)`. `get()` then returns `None` and `item.metadata` raises an `AttributeError` - which `_watch_resource_loop` treats as a fatal error and terminates the sidecar. Not good.
Fix:
    ```
    for key in relevant_keys - exist_keys:
        item = resource_objects.get(key)  # Use snapshot, not live dict
        if item is None:
            continue  # another thread has already removed the key
        metadata = item.metadata
    
    ```

2. Pre-existing: `files_changed` for secrets is overwritten instead of `OR` linked

`src/resources.py:204` (unchanged, but passed through by the PR):
```
if resource == RESOURCE_CONFIGMAP:
    files_changed |= _process_config_map(...)   # correct
else:
    files_changed = _process_secret(...)         # <- assignment instead of |=
```
If there are multiple secrets to be removed: If the first one returns `True` and the second returns `False`, `files_changed` will be set to `False` at the end - the downstream script and webhook calls will not be triggered, even though files were deleted. This was not introduced by this PR, but it affects precisely the cleanup path that the PR now executes more frequently (all namespace-specific stale keys).

Conclusion: Finding 1 is the core of the original change request - the `.copy()` is in the right place, but the loop body must also be changed to `resource_objects`. Finding 2 is a pre-existing `|=`-bug that is mentioned for the sake of completeness.

CarstenSon:
Hi @ChristianGeie: thanks a lot for the thorough review! I didn't catch the |= issue and missed the 2nd occurence of the _resources_object_map before.. 

I included your comments now and also fixed the |= 

Hope I didn't miss anything further :grinning:

src/resources.py
```diff
@@ -189,19 +189,26 @@ def list_resources(label, label_value, target_folder, request_url, request_metho
...
+    # Clear the cache that is not listed. Scope the diff to this namespace: the cache is shared across per-namespace threads, so an unscoped diff would let one thread delete another's resources. 
+    resource_objects = _resources_object_map[resource].copy()
+    relevant_keys = {
+        key for key, item in resource_objects.items()
+        if namespace == "ALL" or item.metadata.namespace == namespace
+    }
+    for key in relevant_keys - exist_keys:
+        item = resource_objects.get(key)
+        if item is None:
+          continue  # another thread has already removed the key
```
ChristianGeie: One last nit: the `continue` inside the `if item is None` block uses 2-space indentation instead of the 4-space convention used throughout the rest of the file.

ChristianGeie: All three concerns are now addressed - thank you! Aside from intention issue above, the logic is correct and the fix is clean. Happy to approve once this is adjusted.

CarstenSon:
@ChristianGeie: I think we did it now :smile: 

thanks a lot for the thorough reivew - looking forward to use this patch in production then!

ChristianGeie (approved):
lgtm

PR #576, opened 2026-05-12
Add the missing build dependencies for all platforms, not just ARMv7.

ChristianGeie:
This ensures that the build process works correctly on all target platforms.

  - include musl-dev and libffi-dev for all platforms:
    - musl-dev - C library headers and linker support
    - libffi-dev - FFI (Foreign Function Interface) development files


ChristianGeie:
fix #574

PR #585, opened 2026-06-09
Use ENTRYPOINT instead of CMD to allow CLI flags

baurmatt:
Using CMD doesn't allow to pass CLI flags like `--req-password-file`, ENTRYPOINT does :)

### CMD
```bash
$ docker run k8s-sidecar --req-password-file /vault/secrets/admin-password
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "--req-password-file": executable file not found in $PATH: unknown

Run 'docker run --help' for more information
```

### ENTRYPOINT
```bash
$ docker run k8s-sidecar --req-password-file /vault/secrets/admin-password
{"time": "2026-06-09T14:23:39.433191+00:00", "level": "INFO", "msg": "Starting collector"}
{"time": "2026-06-09T14:23:39.433793+00:00", "level": "INFO", "msg": "No folder annotation was provided, defaulting to k8s-sidecar-target-directory"}
{"time": "2026-06-09T14:23:39.433896+00:00", "level": "CRITICAL", "msg": "Should have added {LABEL} as environment variable! Exit"}
```

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -269,6 +269,7 @@ jobs:
         run: |
           mkdir /tmp/sidecar
           mkdir /tmp/sidecar-5xx
+          mkdir /tmp/sidecar-basicauth-args
           echo "Downloading resource files from sidecar..."
           kubectl cp sidecar:/tmp/hello.world /tmp/sidecar/hello.world
           kubectl cp sidecar:/tmp/cm-kubelogo.png /tmp/sidecar/cm-kubelogo.png
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,4 +28,4 @@ ENV PATH="/app/.venv/bin:$PATH"
 # Use the nobody user's numeric UID/GID to satisfy MustRunAsNonRoot PodSecurityPolicies
 # https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups
 USER        65534:65534
-CMD         [ "python", "-u", "-m", "sidecar" ]
+ENTRYPOINT  [ "python", "-u", "-m", "sidecar" ]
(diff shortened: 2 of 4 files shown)


ChristianGeie (changes requested):
The existing test overrides the complete container command, so it bypasses the new image `ENTRYPOINT` and would also pass with the old `CMD`.
Could you use Kubernetes `args` instead? This makes the test fail with the old Dockerfile and verifies the behavior introduced by your PR.


test/resources/sidecar.yaml:
replace existing `command:` line with `args:`

```
metadata:
  name: sidecar-basicauth-args
spec:
  containers:
  - name: sidecar
    args:
      - --req-username-file=/opt/creds/username
      - --req-password-file=/opt/creds/password
```

baurmatt:
Oh good idea, thanks! :) Fixed.

ChristianGeie:
@baurmatt thx.

2 more findings:

1. The CI step at line 289 copies from the `sidecar-basicauth-args` pod without first creating the destination directory:

```
# lines 270–271 create the other dirs, but not this one:
mkdir /tmp/sidecar
mkdir /tmp/sidecar-5xx
```

```
# line 289 — destination directory does not exist yet:
kubectl cp sidecar-basicauth-args:/tmp/secured.txt /tmp/sidecar-basicauth-args/secured.txt
```

`kubectl cp` will fail with "no such file or directory", causing the content check at line 389 (check_content "allowed" `/tmp/sidecar-basicauth-args/secured.txt`) to silently fail as well. This pre-dates your PR, but since the PR specifically exercises this test path, a `mkdir /tmp/sidecar-basicauth-args` alongside the others at line 271 would be a clean fix to include here.

2. The "Configuration CLI Flags" section documents `--req-username-file` and `--req-password-file` but doesn't show how to pass them in a Kubernetes manifest. Before this PR, users would have guessed command: (which happened to work because it replaced the full invocation). With `ENTRYPOINT` now set, the correct pattern is `args:` - but that's not obvious without an example. A small snippet like the following would make the migration path clear:

```
containers:
  - name: k8s-sidecar
    image: ghcr.io/kiwigrid/k8s-sidecar:latest
    args:
      - --req-username-file=/path/to/username
      - --req-password-file=/path/to/password
```

thx a lot

baurmatt:
@ChristianGeie Thanks, done! :)

ChristianGeie (approved):
lgtm

ChristianGeie:
Closed and re-open to trigger tests against current `master` branch

PR #603, opened 2026-07-14
feat: add FOLDER_PER_NAMESPACE to organize output files by namespace …

### LINKED ISSUE #602: Support namespacing output files with FOLDER_PER_NAMESPACE
## Problem

When the sidecar watches resources across multiple namespaces (e.g. `NAMESPACE=ALL` or a comma-separated list), all matching ConfigMaps/Secrets are written flat into the same `FOLDER`, regardless of which namespace they came from. There's no way to organize the output files by namespace, which makes it harder to tell at a glance which namespace a given file originated from, especially in setups with many namespaces feeding into a shared mount.

## Proposed solution

Add a new environment variable, `FOLDER_PER_NAMESPACE`, that when set to `true` places each resource's files under a subfolder of `FOLDER` named after the resource's namespace, for better organization:

- No `FOLDER_ANNOTATION` override → `FOLDER/<namespace>/<file>`
- Relative `FOLDER_ANNOTATION` override → `FOLDER/<namespace>/<annotation>/<file>`
- Absolute `FOLDER_ANNOTATION` override → unchanged (absolute paths bypass namespacing, since the user is explicitly opting out of `FOLDER`-relative placement)

Default is `false`, so existing setups are unaffected.

As a side effect, this also avoids filename collisions between namespaces (e.g. two ConfigMaps in different namespaces both mounting `values.yaml`), but the main goal is keeping output organized per namespace.

## Example

With `FOLDER=/tmp/`, `FOLDER_PER_NAMESPACE=true`, and a ConfigMap named `app-config` in namespace `team-a` and another in `team-b`:

- `/tmp/team-a/values.yaml`
- `/tmp/team-b/values.yaml`

I have an implementation ready and will open a PR referencing this issue.

sasi4srinivas:
#602 

## Problem

When the sidecar watches resources across multiple namespaces (e.g. `NAMESPACE=ALL` or a comma-separated list), all matching ConfigMaps/Secrets are written flat into the same `FOLDER`, regardless of which namespace they came from. There's no way to organize the output files by namespace, which makes it harder to tell at a glance which namespace a given file originated from, especially in setups with many namespaces feeding into a shared mount.

## Proposed solution

Add a new environment variable, `FOLDER_PER_NAMESPACE`, that when set to `true` places each resource's files under a subfolder of `FOLDER` named after the resource's namespace, for better organization:

- No `FOLDER_ANNOTATION` override → `FOLDER/<namespace>/<file>`
- Relative `FOLDER_ANNOTATION` override → `FOLDER/<namespace>/<annotation>/<file>`
- Absolute `FOLDER_ANNOTATION` override → unchanged (absolute paths bypass namespacing, since the user is explicitly opting out of `FOLDER`-relative placement)

Default is `false`, so existing setups are unaffected.

As a side effect, this also avoids filename collisions between namespaces (e.g. two ConfigMaps in different namespaces both mounting `values.yaml`), but the main goal is keeping output organized per namespace.

## Example

With `FOLDER=/tmp/`, `FOLDER_PER_NAMESPACE=true`, and a ConfigMap named `app-config` in namespace `team-a` and another in `team-b`:

- `/tmp/team-a/values.yaml`
- `/tmp/team-b/values.yaml`

I have an implementation ready and will open a PR referencing this issue.


ChristianGeie:
Thanks for the PR @sasi4srinivas

The implementation is clean and matches exactly the behavior described in #602. A few comments:

The logic in `_get_destination_folder` (`src/resources.py`) correctly handles all four cases:

- No annotation → `FOLDER/<namespace>/`
- Relative `FOLDER_ANNOTATION` → `FOLDER/<namespace>/<annotation>/`
- Absolute `FOLDER_ANNOTATION` → unchanged; namespace is ignored
- `FOLDER_PER_NAMESPACE=false` (default) → exact behavior as before; no regression

Directory creation is straightforward, since `write_data_to_file` (`src/helpers.py`) already creates directories recursively using `os.makedirs` - no additional adjustments were necessary for the new nested namespace subfolders.

Good E2E coverage: new `sidecar-folder-per-namespace` pod with `NAMESPACE=ALL`, new `folder-per-ns-test` test namespace, verification of all four path variants, including file content diffs.

I found 2 non blocking nits:
Trailing whitespace on two blank lines: in `src/resources.py` after `return dest_folder` and in `src/sidecar.py` after `return -1`. This is purely cosmetic, but my inner monk is triggered ;)

src/resources.py
```diff
@@ -92,16 +92,23 @@ def _get_file_data_and_name(full_filename, content, enable_5xx, content_type=CON
...
         else:
-            dest_folder = os.path.join(default_folder, folder_annotation)
+            if folder_per_namespace:
+                dest_folder = os.path.join(default_folder, metadata.namespace, folder_annotation)
+            else:
+                dest_folder = os.path.join(default_folder, folder_annotation)
         logger.info(f"Found a folder override annotation, "
                     f"placing the {metadata.name} in: {dest_folder}")
         return dest_folder
+
```
ChristianGeie: Cosmetic: trailing whitespaces

src/sidecar.py
```diff
@@ -64,6 +65,8 @@ def main():
     if target_folder is None:
         logger.fatal(f"Should have added {FOLDER} as environment variable! Exit")
         return -1
+
```
ChristianGeie: Cosmetic: trailing whitespaces

ChristianGeie (changes requested):

src/resources.py
sasi4srinivas: @ChristianGeie I push a commit to fix this.

src/sidecar.py
sasi4srinivas: @ChristianGeie  I pushed a change to fix this.

ChristianGeie:
@sasi4srinivas 
The build is not red because of a bug in the sidecar logic, but because of an error in the new test step itself.

The ConfigMap in resources.yaml defines the content:
```
data:
  hello-fpn.world: |-
    Hello FPN Default!
```
`|-` removes the final line break
This is fine like all other Test-fixtures.

The problem lies in the comparison operator: The existing helper function `check_content()` in `build_and_test.yaml` (line 107) uses `echo -n` specifically to avoid this trailing newline:

```
check_content() {
  echo -n "$1" | diff - "$2" > /dev/null || { echo "$2 mismatch"; return 1; }
}
```

However, the new step in this PR uses `echo` without the `-n` option:

```
kubectl exec sidecar-folder-per-namespace -- sh -c "cat /tmp/default/hello-fpn.world" | diff - <(echo "Hello FPN Default!") || ...
```

**Fix:**
In all 4 `diff - <(echo "...")` calls in the "Verify FOLDER_PER_NAMESPACE behavior" step (`build_and_test.yaml`), replace `echo` with `echo -n`, consistent with `check_content()`:

```
kubectl exec sidecar-folder-per-namespace -- sh -c "cat /tmp/default/hello-fpn.world" | diff - <(echo -n "Hello FPN Default!") || { echo "hello-fpn.world content mismatch"; exit 1; }
```

...and similarly for the other three lines (relative-fpn.txt, absolute-fpn.txt, hello-fpn-others.world)

ChristianGeie (approved):
lgtm

Issue #338, opened 2024-03-23
sync stop working on k8s api errors, liveness check needed?


fctb:
When we get k8s api errors the sync stop working silently 
like
`calling kubernetes: (410)
Reason: Expired: The resourceVersion for the provided watch is too old.
`
After on debuglevel you only see the msg:
`Performing watch-based sync on secret resources: {'label_selector': 'grafana_dashboard_v10=1', 'timeout_seconds': '300', '_request_timeout': '330'}`
the msg for configmap stops:
`Performing watch-based sync on configmap resources: {'label_selector': 'grafana_dashboard_v10=1', 'timeout_seconds': '300', '_request_timeout': '330'}`
as well as other debug messages related to configmap. We only have matching configmaps in this cluster.

It looks like that the process for configmap are dead. The process itself is still there.

Make it sense to introduce a liveness check (dead man switch like), that on problems the hole container get restartet?

container yaml:
```
  - env:
    - name: REQ_TIMEOUT
      value: "60"
    - name: IGNORE_ALREADY_PROCESSED
      value: "true"
    - name: METHOD
      value: WATCH
    - name: LABEL
      value: grafana_dashboard_v10
    - name: LABEL_VALUE
      value: "1"
    - name: LOG_LEVEL
      value: debug
    - name: FOLDER
      value: /tmp/dashboards
    - name: RESOURCE
      value: both
    - name: NAMESPACE
      value: ALL
    - name: REQ_USERNAME
      valueFrom:
        secretKeyRef:
          key: admin-user
          name: grafana-admin-password
    - name: REQ_PASSWORD
      valueFrom:
        secretKeyRef:
          key: admin-password
          name: grafana-admin-password
    - name: REQ_URL
      value: http://localhost:3000/api/admin/provisioning/dashboards/reload
    - name: REQ_METHOD
      value: POST
    - name: WATCH_SERVER_TIMEOUT
      value: "300"
    - name: WATCH_CLIENT_TIMEOUT
      value: "330"
    image: quay.io/kiwigrid/k8s-sidecar:1.26.1
    imagePullPolicy: IfNotPresent
    name: grafana-sc-dashboard
    resources: {}
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      seccompProfile:
        type: RuntimeDefault
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
[...]

PR #626, opened 2026-09-01
fix: detect a stalled watch by tracking K8s contact per watcher thread

### LINKED ISSUE #621: WATCH mode can go silent indefinitely, and /healthz cannot detect it — the liveness stamp is refreshed unconditionally by the parent process
Related to #338 (closed 2025-08-15 by referencing #417). I believe #338 is not actually fixed and would suggest reopening it; this report adds the missing root cause analysis plus a live measurement showing the health endpoint returning `200 OK` while the watch had been dead for eight hours.

## Summary

In `METHOD=WATCH`, a watcher child process can block forever inside the event stream. The process stays alive, the container keeps running with 0 restarts, readiness stays `True`, and **nothing is logged** — the sidecar simply stops delivering resources. This is the exact condition described in #338 ("sync stop working silently … the process itself is still there") and independently confirmed there by a second user in December 2024.

The health endpoint added in #417 was expected to resolve this. It cannot, by construction: both of its liveness criteria are unable to observe a stalled watch stream.

## Impact

We hit this three times in two weeks on the same workload (Loki ruler rules, delivered by this sidecar). Because the consumer keeps evaluating whatever is on disk, the effect is silent divergence rather than an outage: one replica served rules that no longer existed in Git.

| Date | Duration until noticed | How it was noticed |
|---|---|---|
| 2024-03-23 | — | #338, reported by another user |
| 2026-08-14 | ~2 days | by accident, an alert fired from a rule version no longer in Git |
| 2026-08-26 | ~8 hours | by accident, one replica held 5 of 6 rule groups |
| 2026-08-28 | ~3h45 | custom log-silence alert we had to build ourselves |

On 2026-08-26 we verified the health endpoint directly while the watch had been dead for eight hours: `/healthz` answered **`200 OK`**, `restartCount` was 0, `ready` was `true`. The liveness probe was correctly configured and live (verified in the pod manifest).

The last log line before going silent, in all three of our cases, was an ordinary `Loading incluster config...` — no exception, no traceback, no `ApiException`. Note this differs from #338, where a visible `410 Expired: resourceVersion too old` preceded the stall. Same end state, likely different trigger.

## Root cause

Two independent defects. Line numbers are from `2.10.1`.

### 1. `last_k8s_contact` is stamped unconditionally by the parent process

`healthz.py` treats `last_k8s_contact` as evidence that the API server was reached:

```python
K8S_CONTACT_THRESHOLD_SECONDS = 60  # tolerated delay before declaring not live
...
elif (now - last_k8s_contact) > timedelta(seconds=K8S_CONTACT_THRESHOLD_SECONDS):
    status = 503
    body = "NOT LIVE (K8s contact lost)"
```

`update_k8s_contact()` is called from two places, and they contradict each other.

The **child** process stamps it per received event (`resources.py`, ~line 398) — that is the correct signal, and its comment says so:

```python
update_k8s_contact()  # To be sure that every event received is counted as "K8s alive"
```

But the **parent** process stamps the very same variable every 5 seconds, unconditionally, in `watch_for_changes()` (`resources.py`, ~line 481):

```python
while True:
    # Update k8s contact timestamp to show the main process is alive and watchers are running
    update_k8s_contact()
    died = False
    for proc, ns, resource in processes:
        if not proc.is_alive():
            ...
    sleep(5)
```

The parent's stamp permanently overwrites the child's. `last_k8s_contact` therefore means "the parent loop is running", never "the API server was reached", and the 60-second threshold can never expire while the container is up. The dead man's switch requested in #338 is present in form, but it is operated from a place that knows nothing about the watch.

### 2. The watcher liveness check is only `proc.is_alive()`

```python
elif watcher_processes and not all(p.is_alive() for p in watcher_processes):
    status = 503
```

The child process *is* alive. It is blocked in `for event in stream:` inside `_watch_resource_iterator()`. `is_alive()` only becomes `False` if the process actually exits, which is precisely what does not happen here. #338 already stated this ("The process itself is still there"), and the implemented check does not address it.

### Note: the client-side timeout does not prevent the stall

`_watch_resource_iterator()` (`resources.py`, ~line 373) passes both timeouts correctly:

```python
additional_args = {
    'label_selector': label_selector,
    'timeout_seconds': WATCH_SERVER_TIMEOUT,     # 60 in our config
    '_request_timeout': WATCH_CLIENT_TIMEOUT,    # 66 in our config
}
stream = watch.Watch().stream(getattr(v1, _list_namespace[namespace][resource]), **additional_args)
for event in stream:      # blocked here for 4h / 8h / 2 days
```

We ran with `WATCH_CLIENT_TIMEOUT` (66) deliberately greater than `WATCH_SERVER_TIMEOUT` (60), as recommended. The stream still hung for hours. So this is not a misconfiguration of the documented knobs — `_request_timeout` does not reliably terminate a stalled streaming watch, which is why an external watchdog is needed rather than tighter timeout tuning.

## Detection (in case it helps others)

Since neither the probe, the restart count, nor readiness reflect the condition, the only reliable signal we found is **log silence**. A healthy sidecar logs at a fixed cadence (one line per watch reconnect, i.e. per `WATCH_SERVER_TIMEOUT`); a stalled one logs exactly nothing. Note that a dead sidecar produces *no time series at all* rather than a zero, so any alert on this must count healthy pods and compare against the expected replica count, not test for a low per-pod value.

## Workaround

`METHOD=SLEEP` avoids the defect entirely: there is no long-lived stream that can stall, and a failed LIST goes through the `except` branch and is retried on the next iteration. We switched to `SLEEP` with `SLEEP_TIME=60` and it has been stable since. This is a workaround, not a fix — defect 1 means a stalled LIST loop would be equally invisible.

## Suggested fix

The cleanest change is to make `last_k8s_contact` mean what `healthz.py` already assumes it means:

1. **Remove the unconditional `update_k8s_contact()` from the parent loop in `watch_for_changes()`.** It is the single reason the threshold can never expire.
2. **Stamp it in the child on every watch (re)connect**, not only per event — i.e. right after `watch.Watch().stream(...)` returns and at the top of each `_watch_resource_loop` iteration. Reconnects happen every `WATCH_SERVER_TIMEOUT` regardless of cluster activity, so this gives a genuine heartbeat on an idle cluster while still expiring when the stream stalls. The same stamp in the `SLEEP` branch after each successful `list_resources()` would cover `METHOD=SLEEP` too.
3. **Require `K8S_CONTACT_THRESHOLD_SECONDS > WATCH_SERVER_TIMEOUT`.** Both currently default to 60, so even after (1) and (2) the check would sit exactly on the reconnect boundary and flap. A threshold derived from the server timeout (e.g. `2 × WATCH_SERVER_TIMEOUT`) or a documented constraint between the two would avoid that.

If preferred, an equivalent alternative is a shared `multiprocessing.Value` per watcher holding the last successful loop timestamp, evaluated by the parent — but that duplicates what `last_k8s_contact` was already meant to be.

I am happy to open a PR for this if the approach sounds right.

## Environment

- `kiwigrid/k8s-sidecar:2.10.1` (linux/arm64, Alpine/musl, Python 3.15.0b2)
- `kubernetes` python client 32.0.1, `urllib3` 2.7.0
- Kubernetes v1.36.3 (k3s)
- Deployed via the `grafana/loki` Helm chart 7.3.0 as the ruler rules sidecar
- `METHOD=WATCH`, `RESOURCE=both`, `NAMESPACE=ALL`, `LABEL=loki_rule`, `WATCH_SERVER_TIMEOUT=60`, `WATCH_CLIENT_TIMEOUT=66`, `LOG_LEVEL=INFO`
- livenessProbe on `/healthz:8080`, `periodSeconds: 30`, `failureThreshold: 4`

thomas-baumeister:
Fixes #621, and the condition originally reported in #338.

Thanks @ChristianGeie for verifying both root causes against master, and for the correction on threads vs. processes. The terminology in #621 was wrong, and it turns out to matter for the fix; see below.

## What was wrong

A watcher thread can block indefinitely inside `for event in stream:` when a watch stream stalls. The thread stays alive, so `is_alive()` reports healthy, and until now the contact timestamp did too, because `watch_for_changes()` refreshed a single shared `last_k8s_contact` every 5 seconds from the main loop, unconditionally. That only ever proved that the main loop was running, which is always true while the container is up. So `/healthz` answered `200 OK` while the sidecar had silently stopped delivering resources.

## What this changes

**1. The main loop no longer stamps.** Only watchers report their own contact.

**2. Watchers stamp on every completed watch reconnect** (the iterator returning proves the stream ran its full course) **and after every completed LIST in SLEEP mode.** The pre-existing per-event stamp is kept, but it is not sufficient on its own: an idle cluster produces no events, so without the reconnect stamp a perfectly healthy sidecar would age out and be declared not live.

**3. The threshold is derived from how often watchers actually report in:** `2 × WATCH_SERVER_TIMEOUT`, or `2 × SLEEP_TIME` when polling. It was previously hardcoded to 60, exactly equal to the default `WATCH_SERVER_TIMEOUT`, so once the check could fail at all it would have sat right on the reconnect boundary and flapped. `K8S_CONTACT_THRESHOLD_SECONDS` still overrides it, so anyone pinning the old value keeps it.

## One addition beyond the three points discussed in the issue

The heartbeat is now tracked **per watcher thread** (keyed by `threading.get_ident()`) rather than in one shared variable.

This is precisely where the threads-not-processes correction becomes load-bearing. With `RESOURCE=both` (the default, and what both reports ran) there is one watcher per resource kind, and they share the module global directly. A healthy watcher therefore keeps refreshing the shared timestamp while its sibling sits on a stalled stream, and the check still never fires. Moving the stamp out of the main loop alone would have fixed the reported symptom only when *every* watcher stalls at once, which is the lucky case, not the general one.

## Verification

I could not find an existing unit test layer to hang this off, since `test/` is the kind-based e2e setup. So I verified it by simulating the failure directly: two live watcher threads, one of which stops reporting contact while remaining alive, against a real health server.

| Variant | `/healthz` after threshold |
|---|---|
| before this change, main loop stamping | `200 OK`, the reported defect |
| before this change, watchers stamping (the three issue points only) | `200 OK`, healthy sibling masks the stalled one |
| after this change | `503 NOT LIVE (K8s contact lost)` |

`is_alive()` is `True` for the stalled watcher in all three runs, which is why the existing thread check cannot catch this.

A second check covers `resources.py` rather than `healthz.py`: it drives `_watch_resource_loop()` through one iteration in each mode with the API calls stubbed out, and asserts that a heartbeat is recorded in both the WATCH and the SLEEP branch, and that `watch_for_changes()` records none. I added it after the first version of this branch turned out to carry the comment describing the reconnect stamp without the call itself, which the health-server simulation could not see because it never runs `resources.py`.

Happy to fold those simulations into the repo as tests if you want unit tests introduced. It needs no cluster, only `threading` and the health server, but it would mean adding a test runner and a CI job, so I left that out of this PR rather than deciding it for you.

## Notes

- `METHOD=LIST` is unaffected: with no watchers registered, `_stale_watchers()` is empty and a one-shot run is never reported as not live.
- Behaviour on an idle cluster is now *better* defined than before: the reconnect stamp gives a real heartbeat where previously only the (unfalsifiable) main-loop stamp existed.
- The failure this addresses is invisible from inside the container by any other means we could find. For anyone hitting it before this ships, `METHOD=SLEEP` avoids the stall entirely, and log silence is the only external signal: a healthy sidecar logs once per reconnect, a stalled one logs nothing at all.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -7,6 +7,24 @@ on:
   - pull_request
   - workflow_dispatch
 jobs:
+  # Unit tests for the liveness signal (#621, #626). Standard library only, so
+  # this needs no test runner dependency and finishes in seconds; it runs
+  # independently of the image build. `pip install .` is only there to pull the
+  # runtime dependencies that src/ imports.
+  unit-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+      - name: Install runtime dependencies
+        run: pip install .
+      - name: Run unit tests
+        run: python -m unittest discover -s test/unit -t test/unit
+
   build:
     runs-on: ubuntu-latest
     steps:
--- a/src/healthz.py
+++ b/src/healthz.py
@@ -7,17 +7,25 @@ import threading
 from datetime import datetime, timedelta, timezone
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from threading import Thread
-from typing import List
+from typing import Dict, List, Optional
 
 from logger import get_log_config
 
 # Health state variables
 is_ready = False
-last_k8s_contact = datetime.now(timezone.utc)
 watcher_processes: List[Thread] = []
 
-# Settings
-K8S_CONTACT_THRESHOLD_SECONDS = 60  # tolerated delay before declaring not live
+# Last successful Kubernetes contact, tracked PER WATCHER THREAD (keyed by
+# thread ident). A single shared timestamp is not enough: with RESOURCE=both
+# there is one watcher per resource kind, and a healthy one would keep the
+# shared timestamp fresh while another sits on a stalled stream (see #621).
+watcher_heartbeats: Dict[int, datetime] = {}
+_heartbeat_lock = threading.Lock()
+
+# Tolerated delay before declaring not live. Overridden in
+# register_watcher_processes() based on how often watchers actually report in;
+# the default only applies until watchers are registered.
+K8S_CONTACT_THRESHOLD_SECONDS = 60
 
 
 class HealthCheckFilter(logging.Filter):
@@ -32,7 +40,7 @@ class HealthHandler(BaseHTTPRequestHandler):
     server_version = "HealthHTTP/1.0"
 
     def do_GET(self):
-        global is_ready, last_k8s_contact, watcher_processes
+        global is_ready, watcher_processes
 
         if self.path != "/healthz":
             self.send_response(404)
@@ -49,16 +57,16 @@ class HealthHandler(BaseHTTPRequestHandler):
         if not is_ready:
             status = 503
             body = "NOT READY"
-        # Liveness check (k8s contact)
-        elif (now - last_k8s_contact) > timedelta(
-                seconds=K8S_CONTACT_THRESHOLD_SECONDS
-        ):
-            status = 503
-            body = "NOT LIVE (K8s contact lost)"
-        # Liveness check (watcher processes)
+        # Liveness check (watcher threads still running)
         elif watcher_processes and not all(p.is_alive() for p in watcher_processes):
             status = 503
             body = "NOT LIVE (watcher thread died)"
+        # Liveness check (k8s contact, per watcher). This is what catches a
+        # watcher that is still alive but stuck in a stream that never returns
+        # -- the failure mode of #338/#621, which is_alive() cannot see.
+        elif _stale_watchers(now):
+            status = 503
+            body = "NOT LIVE (K8s contact lost)"
         else:
             status = 200
             body = "OK"
@@ -95,17 +103,57 @@ def mark_ready():
 
 def update_k8s_contact():
     """
-    Update the timestamp of the last successful Kubernetes contact.
-    """
-    global last_k8s_contact
-    last_k8s_contact = datetime.now(timezone.utc)
+    Record a successful Kubernetes contact for the CALLING watcher thread.
 
-def register_watcher_processes(processes: List[Thread]):
+    Must be called from the watcher thread itself -- the heartbeat is keyed by
+    threading.get_ident(). Calling it from the main thread would only refresh
+    an entry nothing looks at, which was the original defect (#621): the main
+    loop refreshed a single shared timestamp every 5s regardless of whether any
+    watcher had reached the API server, so the liveness check could never fail.
     """
-    Register the list of watcher threads to be monitored for liveness.
+    with _heartbeat_lock:
+        watcher_heartbeats[threading.get_ident()] = datetime.now(timezone.utc)
+
+
+def _stale_watchers(now: datetime) -> List[Thread]:
     """
-    global watcher_processes
+    Return the registered watcher threads whose last Kubernetes contact is older
+    than the tolerated threshold. Empty while no watchers are registered (e.g.
+    METHOD=LIST), so a one-shot run is never reported as not live.
+    """
+    threshold = timedelta(seconds=K8S_CONTACT_THRESHOLD_SECONDS)
+    with _heartbeat_lock:
+        return [
+            t for t in watcher_processes
+            if (now - watcher_heartbeats.get(t.ident, now)) > threshold
+        ]
+
+
+def register_watcher_processes(processes: List[Thread], heartbeat_interval: Optional[int] = None):
+    """
+    Register the watcher threads to be monitored for liveness and seed their
+    heartbeats, so the clock starts at registration rather than at first report.
+
+    heartbeat_interval is how often a healthy watcher is expected to report in
+    (WATCH_SERVER_TIMEOUT when watching, SLEEP_TIME when polling). The threshold
+    is derived as twice that, so the check never sits exactly on the reconnect
+    boundary and flaps. K8S_CONTACT_THRESHOLD_SECONDS in the environment
+    overrides the derived value.
+    """
+    global watcher_processes, K8S_CONTACT_THRESHOLD_SECONDS
     watcher_processes = processes
+
+    override = os.getenv("K8S_CONTACT_THRESHOLD_SECONDS")
+    if override:
+        K8S_CONTACT_THRESHOLD_SECONDS = int(override)
+    elif heartbeat_interval:
+        K8S_CONTACT_THRESHOLD_SECONDS = 2 * int(heartbeat_interval)
+
+    now = datetime.now(timezone.utc)
+    with _heartbeat_lock:
+        watcher_heartbeats.clear()
+        for thread in processes:
+            watcher_heartbeats[thread.ident] = now
 
 def _create_health_http_server(health_port: int) -> ThreadingHTTPServer:
     """
--- a/src/resources.py
+++ b/src/resources.py
@@ -448,11 +448,21 @@ def _watch_resource_loop(shutdown_event, mode, label, label_value, target_folder
                 list_resources(label, label_value, target_folder, request_url, request_method, request_payload,
                                namespace, folder_annotation, resource, unique_filenames, script, enable_5xx,
                                ignore_already_processed, resource_name, folder_per_namespace)
+                # A completed LIST is proof of contact for this watcher.
+                update_k8s_contact()
                 sleep(int(os.getenv("SLEEP_TIME", 60)))
             else:
                 _watch_resource_iterator(label, label_value, target_folder, request_url, request_method, request_payload,
                                          namespace, folder_annotation, resource, unique_filenames, script, enable_5xx,
                                          ignore_already_processed, folder_per_namespace)
+                # The iterator returns once the server closes the stream after
+                # WATCH_SERVER_TIMEOUT, so reaching this line proves the
+                # connection ran its full course. This is the heartbeat that
+                # keeps an IDLE cluster healthy -- the per-event stamp inside
+                # the iterator never fires when no resources change. A stalled
+                # stream never returns here, so the heartbeat ages out and
+                # /healthz reports 503 (#338, #621).
+                update_k8s_contact()
         except ApiException as e:
             if e.status != 500:
                 logger.error(f"ApiException when calling kubernetes: {e}\n")
@@ -472,6 +482,23 @@ def _watch_resource_loop(shutdown_event, mode, label, label_value, target_folder
     logger.info(f"Shutdown event received, stopping watcher for {namespace}/{resource}.")
 
 
+def heartbeat_interval(mode, namespace, resource_name):
+    """
+    How often a healthy watcher is expected to report a successful Kubernetes
+    contact. The liveness threshold in healthz is derived from this value.
+
+    The predicate has to match the branch taken in _watch_resource_loop()
+    exactly, not just `mode`: with RESOURCE_NAME set on a non-ALL namespace,
+    WATCH also runs the list-and-sleep path and therefore reports in at
+    SLEEP_TIME rather than at WATCH_SERVER_TIMEOUT. Deriving from `mode` alone
+    sizes the threshold against the wrong interval and flaps as soon as
+    SLEEP_TIME exceeds twice WATCH_SERVER_TIMEOUT.
+    """
+    if mode == "SLEEP" or (namespace != 'ALL' and resource_name):
+        return int(os.getenv("SLEEP_TIME", 60))
+    return int(WATCH_SERVER_TIMEOUT)
+
+
 def watch_for_changes(mode, label, label_value, target_folder, request_url, request_method, request_payload,
                       current_namespace, folder_annotation, resources, unique_filenames, script, enable_5xx,
                       ignore_already_processed, resource_name, folder_per_namespace):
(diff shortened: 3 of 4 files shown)


ChristianGeie:
@thomas-baumeister THx for the PR.

The healthz logic is clean, but your PR exposes an existing, serious bug in `resources.py` that renders part of your fix ineffective - i need to fix this before the merge.

PR-Review:
The per-thread heartbeat approach (`watcher_heartbeats: Dict[int, datetime]`, keyed by `threading.get_ident()`) actually resolves both root causes mentioned in the issue #621, plus the additional bug that you found himself (the shared timestamp when `RESOURCE=both` - a healthy watcher would have masked the other one). Lock handling for `watcher_heartbeats`/ `watcher_processes` is correct, `_stale_watchers()` remains empty when `METHOD=LIST` (no false-negative liveness for one-shot runs) and the derivation of `K8S_CONTACT_THRESHOLD_SECONDS = 2 × heartbeat_interval` with an override via the env var is exactly what was proposed in the issue.

Critical Issue:
The SLEEP branch calls `list_resources()` with one argument too few.
In `resources.py`, `_watch_resource_loop()`, the line immediately before the new `update_k8s_contact()`:
```
list_resources(label, label_value, target_folder, request_url, request_method, request_payload,
               namespace, folder_annotation, resource, unique_filenames, script, enable_5xx,
               ignore_already_processed, resource_name)
```

`list_resources()` now expects 15 positional arguments (since PR #603, when `folder_per_namespace` was added)-but only 14 are passed here.
This isn't a bug in your PR; it's already present in the master branch (Python consistently throws a `TypeError: "missing 1 required positional argument: 'folder_per_namespace'"`), but your PR inserts its SLEEP heartbeat line right after that-and as a result, that line is never reached.

Consequence:
In `METHOD=SLEEP` (and in WATCH mode with `RESOURCE_NAME` + Namespace ≠ `ALL`, which uses the same branch), each cycle throws a `TypeError` that is caught by the generic `except Exception` block-the thread remains alive but never syncs anything again. This was already broken before (silent sync halt), but your PR turns it into an actively failing liveness check:
The heartbeat for this Watcher continues to age inexorably beyond the registration time, `/healthz` consistently reports `503 NOT LIVE`, and a configured `livenessProbe` causes the pod to enter a restart loop-for SLEEP mode users (the workaround recommended in the issue itself!), this is a step backward compared to the silent freeze that occurred previously.
This also explains why the CI doesn't catch this: While `.github/workflows/build_and_test.yaml` deploys a `sidecar-sleep` pod, it only checks `sidecar` and `sidecar-5xx` (both with `METHOD=WATCH`) for updated files based on `change_resources.yaml` - sidecar-sleep is only checked for its startup log, never for update behavior.

Suggestion:
I filee a separate bug fix PR #628 that this one can then build upon; i pass `folder_per_namespace` to this `list_resources()` call and adjust the E2E tests.

Minor issue:
Heartbeat interval mismatch for WATCH+RESOURCE_NAME+Namespace≠ALL
`heartbeat_interval` in `watch_for_changes()` is selected only based on `mode == "SLEEP"`:
`
heartbeat_interval = int(os.getenv("SLEEP_TIME", 60)) if mode == "SLEEP" else int(WATCH_SERVER_TIMEOUT)
`
However, the actual path per thread depends on `mode == "SLEEP" or (namespace != 'ALL' and resource_name)`. When `METHOD=WATCH` with `RESOURCE_NAME` set and namespace ≠ `ALL`, the thread effectively runs at the LIST+`SLEEP_TIME` interval, but the threshold is calculated based on `WATCH_SERVER_TIMEOUT`. In an asymmetric configuration (e.g., `SLEEP_TIME` significantly greater than `2×WATCH_SERVER_TIMEOUT`), this would lead to false-positive “NOT LIVE” flapping. With the default values (both 60), this does not occur, but there is a gap in the derivation.

A quick note:
The PR description mentions two written verification/test simulations (a health server simulation and a test that runs `_watch_resource_loop()` through both modes)-but the diff only includes `src/healthz.py` and `src/resources.py` (`changed_files: 2`), with no test files. If you already have tests like this, please include them-precisely because a test using unspeced mocks for `list_resources()` might have masked the bug mentioned above instead of catching it.

thomas-baumeister:
Thanks for the thorough review, and for #628. All three points are fair.

**On the SLEEP branch.** I had actually found that one too and filed it as #627 shortly before you opened #628, but I clearly failed to connect it to this PR, which is the more important half of what you are pointing out. You are right that my change turns a silent freeze into an actively failing liveness check for exactly the users who moved to SLEEP because of #621. I will rebase on #628 once it lands rather than duplicating the fix here.

**On the heartbeat interval mismatch.** Correct, and fixed. The derivation now mirrors the branch condition in `_watch_resource_loop()` instead of looking at `mode` alone:

```python
if mode == "SLEEP" or (current_namespace != 'ALL' and resource_name):
    heartbeat_interval = int(os.getenv("SLEEP_TIME", 60))
else:
    heartbeat_interval = int(WATCH_SERVER_TIMEOUT)
```

**On the tests, you were right in a way that stings.** The simulations existed but were not in the diff, and the one covering `_watch_resource_loop()` used a permissive `lambda *a, **k` stand-in for `list_resources()`. It did exactly what you predicted: it swallowed the argument count and hid the very bug you found. Worse, an earlier revision of this branch carried the comment describing the reconnect stamp without the call itself, and the health-server simulation could not see that either because it never runs `resources.py`.

Both are now proper unit tests in `test/unit/test_liveness.py`, standard library only, with `create_autospec` so a wrong argument count raises instead of being absorbed. They cover:

- a healthy watcher must not mask a stalled sibling (the `RESOURCE=both` case)
- `METHOD=LIST` registers no watchers and must never report not live
- the threshold derivation and its env override
- registration seeds the heartbeats, so a watcher that has not reported yet is not immediately stale
- both branches of `_watch_resource_loop()` record a heartbeat, including WATCH with `RESOURCE_NAME` on a single namespace

Verification that they actually bite: against this branch without #628 the two tests touching the list path fail with `TypeError: missing a required argument: 'folder_per_namespace'`. With #628 cherry-picked, all nine pass. So they reproduce the bug you described rather than papering over it.

There is also a `unit-test` job in `build_and_test.yaml`. It runs independently of the image build and needs no test runner dependency, since `unittest` is standard library. `pip install .` is only there to pull the runtime dependencies that `src/` imports. If you would rather pin the interpreter with `actions/setup-python`, say the word and I will add it.

I will hold the push until #628 is merged so this branch does not carry a duplicate of your fix, then rebase and update here.

thomas-baumeister:
Rebased onto master now that #628 is in. The unit tests that previously failed with the missing folder_per_namespace argument are green, all nine pass.

ChristianGeie:
@thomas-baumeister 

PR-Review:
1. Your PR ist rebased on #628 -> :+1: 
2. Heartbeat interval mismatch: The derivation now exactly mirrors the branch condition from `_watch_resource_loop()`. I have verified that `current_namespace` is the correct variable within the scope of `watch_for_changes()` and uses exactly the same predicate as in `_watch_resource_loop()` -> :+1:
3. Tests (`test/unit/test_liveness.py`, new unit test job in CI): Using `create_autospec` instead of permissive lambdas is exactly the right response to my "quick note" point
5. Non Blocking, small Points:
- The new unit-test job does not pin a Python version (it uses the default from `ubuntu-latest`). You yourself offered to add `actions/setup-python` if desired - i would support this, in line with the project’s other pinning requirements
- `HeartbeatIntervalTest.test_watch_with_resource_name_is_sized_against_sleep_time` does not test the behaviour, but instead searches the source code of `resources.py` for the string "current_namespace != 'ALL' and resource_name". Ironically, this is exactly the sort of test that doesn’t actually do anything - it works correctly at the moment, but it would break if someone simply renamed or refactored the code without changing the behaviour and it wouldn’t catch a genuine bug if the condition were present in the text but wired up incorrectly. As it is very diffficult to test `watch_for_changes()` directly due to the `while True` loop at the end, a cleaner solution would be to move the interval calculation into a small, standalone helper function and then actually call that function.

thomas-baumeister:
Both points addressed, thanks.

**Python version.** The `unit-test` job now uses `actions/setup-python` pinned by SHA, with `python-version: "3.12"` to match `requires-python` in `pyproject.toml`.

**The source-grep test.** You are right, and it is the more embarrassing of the two given the context. That test asserted on the text of `resources.py` rather than on behaviour, so it would have broken on a harmless rename and stayed green on a real miswiring. Exactly the failure mode I had just been called out for.

Fixed the way you suggested. The calculation now lives in a small standalone function:

```python
def heartbeat_interval(mode, namespace, resource_name):
    if mode == "SLEEP" or (namespace != 'ALL' and resource_name):
        return int(os.getenv("SLEEP_TIME", 60))
    return int(WATCH_SERVER_TIMEOUT)
```

and `watch_for_changes()` just calls it. That sidesteps the `while True` problem you mention, since the interesting part is now reachable without running the loop at all.

Four real assertions replace the grep, covering each branch including the one the old derivation got wrong:

- `SLEEP` uses `SLEEP_TIME`
- plain `WATCH` uses `WATCH_SERVER_TIMEOUT`
- `WATCH` with `RESOURCE_NAME` on a single namespace uses `SLEEP_TIME`
- `WATCH` with `RESOURCE_NAME` on `ALL` stays on `WATCH_SERVER_TIMEOUT`, since the namespace is what diverts it

Verified that they bite: reverting the derivation to the old `mode`-only form turns the third one red with `AssertionError: 90 != 60`. Twelve tests, all green on the current branch.

ChristianGeie:
Wow, you’re so quick, i can’t keep up with the review at all.
- The interval calculation is now a standalone function and `watch_for_changes()` now actually calls it -> :+1: 
- The four new tests in `HeartbeatIntervalTest` cover all four combinations (`SLEEP`, plain `WATCH`, `WATCH`+`RESOURCE_NAME` on a single namespace, `WATCH`+`RESOURCE_NAME` on `ALL`) and call the function directly - no more text grep -> :+1:

thomas-baumeister:
Actually, you are doing all the work and pointing out my mistakes. I just need to fix them ;-)

ChristianGeie (approved):
ltgm

